PR #2, opened 2020-10-06
Add a few API endpoints & bindings

umberto-sonnino:


cpp/src/bindings/bindings_linear_animation_instance.cpp
```diff
@@ -64,10 +60,10 @@ extern "C"
                 enumField = ::pingPongLoopField;
                 break;
             }
+            
+            loopValue = env->GetStaticObjectField(::loopClass, enumField);
         }
 
-        jobject loopValue = env->GetStaticObjectField(::loopClass, enumField);
-
         return loopValue;
```
mjtalbot, line 67: I'm not sure I totally understand this. 

so we have our cpp animation instance, we advance it, and we pass in an argument to it, that it may change 
(any reason we don't return a value from advance btw? presumably advance already returns if its going to keep going?) 

and now everytime the animation loops (either because of pingpong/ loop presumably, how would an animation set to oneshot loop btw? or does coming to its end count as a 'loop' there?) we will return the active loop mode. (so we return null most of the time, but the active loop mode sometimes) 

I'm guessing this is so we can track how many times something has looped. 
I wonder if this should either live in the rive-cpp lib (so it can be shared with ios & wasm) or if it should live in the kotlin side (by checking the time & direction of an animation instance), so its clear that if you want to track this kinda stuff you can implement your own way of tracking 'stuff'

mjtalbot: slight concern here as well is that we're wrapping the advance function (which returns true/false for if its going to continue or not) but we're not exposing its output, which doesn't quite seem right. 

I suppose (and this is guessing) that we get a loop 'oneShot' back if the animation finished, still not sure  why this wouldnt be the signature if this is the signature we want

mjtalbot (approved):
LGTM, with comments

looks like this is something you've probably already discussed with @luigi-rosso though. would love to catch up with that conversation to be on the same page

cpp/src/bindings/bindings_linear_animation_instance.cpp
umberto-sonnino: Here the goal is to extract whether an animation has looped or not.
advance() returns a bool to notify if it advanced, but there wasn't a clear way to know if the animation had looped around, or it pingpong'd. 
So when a loop happens, the flag that is passed in as a parameter gets raised, this is why the code queries for the loop type and reconciles it with the Koktlin enum.

This is indeed living also in the cpp side, as there's a PR pending over there!

As of now, I needed the flag on the Android side to build the example, but this should indeed return the same value as advance. 
We just need to set this up so that we can have both values returned in Kotlin too (a field to be set on the animation instance? An object that wraps both values?)

umberto-sonnino:
> LGTM, with comments
> 
> looks like this is something you've probably already discussed with @luigi-rosso though. would love to catch up with that conversation to be on the same page

Sure, of course! The PR that's pending on the other side is [here](https://github.com/rive-app/rive-cpp/pull/11)

cpp/src/bindings/bindings_linear_animation_instance.cpp
mjtalbot: so is the plan here to made the rive-cpp advance return a combination of 'keepGoing & didLoop' ? 
or is the plan to keep going with the output, plus switching input flags? 

also what is the thinking with the 'loopmode' query here? it totally looks like this should be set by rive-cpp, or it should live in kotlin (if you run advance, and get 'didLoop'. true, i guess you can go and check what the loop mode is, if you want?

umberto-sonnino: Ideally, if we want to have a 1:1 match between the cpp & the binding, we'd have a way of letting the Kotlin layer of both things, the advance return value, and the didLoop value.
If you think switching for the enum on the Kotlin side is cleaner, I can set it up that way!

mjtalbot: i guess it depends on what we want the sig to be. 

personally im leaning towards having advance from the cpp return an object rather than a bool, I guess in this case that would be {keepGoing, didLoop}, and then have our kotlin layer pick out the loop mode. 

or have advance return {keepGoing, didLoop, loopMode} (but i'm guessing that thats not always wanted) 

but i get a feeling there's a reason that I'm not seeing why we didn't approach it that way

mjtalbot: and yeah, i totally agree, would be great for the bindings and the cpp to be a 1:1 match, and any additional customisation beyond that, should really live in the example of a specific use case. 

Otherwise i *guess* the question of, won't all runtimes need this comes up pretty quick, and if thats the case, it should be in the cpp. 

I guess there is some space for things that are platform specific, but ideally our api signature is pretty similar across the board.

umberto-sonnino: The idea was to keep the advance() path as lean as possible since this is being called every frame, and returning an object from a function in cpp might lead to unexpected results.

For example, I had put together a completely different solution on the rive-cpp PR, but passing an output parameter here was found to be the solution with the most minimal impact code-wise. So probably it's best to ask @luigi-rosso directly and see what he thinks about returning the object there.

PR #3, opened 2020-10-06
Android23compat

mjtalbot:


umberto-sonnino (approved):
LGTM, Matrix implementation looks good, we should probably get that in as well since the other call is deprecated

luigi-rosso (approved):
LGTM, can you link me to the canvas? Curious to see the set/get matrix stuff. The decomposing stuff is problematic when there is composed non-uniform scale (which causes skewing). Really similar to our logic here, but it fails when you start really skewing things: https://github.com/rive-app/rive-cpp/blob/88ff17d961fc5cdb37a4e18297595a4e296d01c2/src/math/mat2d.cpp#L77

luigi-rosso:
Tank you.

Issue #8, opened 2020-11-18
Antialiasing and pixelation


applikationsprogramvara:
I tried several riv-files and noticed strange behavior on rendering:
- on Android antialiasing is off - one can clearly see the sharp edges, which are not always required
- on lower SDKs "pixelation" effect appears - strangely enough, due to the fact, that the source is vector. To intensify the effect, I used a "narrow" artboard with width ~400. With wider artboards the effect is lower.

On the web the same riv-file is rendered just OK (version rive-canvas@0.6.5) - with antialiasing and without pixelation.

Here are my questions:
1. How antialiasing can be turned on on Android?
2. It would be great if you take a look into "pixelation" rendering for lower SDKs

Below are results of my experiments:

Source | SDK21 | SDK29 | Web
--- | --- | --- | ---
Overview image. Click to see the full version | <img width="250" alt="portfolio_view" src="https://user-images.githubusercontent.com/38891920/99595264-b75cbf80-29f4-11eb-8a95-faf291eab93a.png"> | <img width="250" alt="portfolio_view" src="https://user-images.githubusercontent.com/38891920/99595272-b9268300-29f4-11eb-96dd-2732ffc27016.png"> | <img width="250" alt="portfolio_view" src="https://user-images.githubusercontent.com/38891920/99595278-ba57b000-29f4-11eb-9056-21389394b435.png">
Closeup circle | ![d1_2](https://user-images.githubusercontent.com/38891920/99595267-b7f55600-29f4-11eb-86cc-eca22899abb7.png) | ![d2_2](https://user-images.githubusercontent.com/38891920/99595275-b9268300-29f4-11eb-833e-cb068bbd801b.png) | ![d3_1](https://user-images.githubusercontent.com/38891920/99595280-baf04680-29f4-11eb-8afc-58c7bfc7ab8a.png)
Closeup text | ![d1_3](https://user-images.githubusercontent.com/38891920/99595271-b88dec80-29f4-11eb-821f-24feaf5561b3.png) | ![d2_3](https://user-images.githubusercontent.com/38891920/99595277-b9bf1980-29f4-11eb-9311-44d55e78dcc6.png) | ![d3_2](https://user-images.githubusercontent.com/38891920/99595282-baf04680-29f4-11eb-9781-6f907b5a2ec7.png)

applikationsprogramvara:
I tried your update from branch antialiasing
https://github.com/rive-app/rive-android/pull/9

It works and looks much better! Thank you!

[...]

PR #9, opened 2020-11-19
Support for Antialiasing

umberto-sonnino:
Enable setting antialiasing as a renderer parameter (defaults to `true`), and this is translated as a static field in JNI.
Updated to the latest rive-cpp & all .so.

Also, adding some logging macros as found [here](https://stackoverflow.com/a/20111129/1706076) to make logging a bit more concise.

Address #8


mjtalbot (approved):
LGTM

PR #13, opened 2020-12-11
update .so files for latest cpp runtime

mjtalbot:
added simple star/polygon check file
updated README with reminders for how to build .so


umberto-sonnino (approved):
LGTM

PR #19, opened 2021-02-14
Android animation view and drawable implementation

MightySeal:
This PR implements android `View` and `Drawable` so it's easier to to use Rive in android development.
The implementation uses Animator which is synchronized with  display frame rate, so it renders at the maximum frame rate, but doesn't exceed it which saves battery life.
Also, some small improvements like code cleanup according to kotlin codestyle, slightly improved example so it focuses on the view usage.
There's also a room for further work and I'm happy to take part in it, so I'll create several issues for the discussion.


umberto-sonnino:
Thanks for taking the time to put this together. We really appreciated your time and effort at Rive - all the contributions we see here are looking nice and clean!

We've been talking about this internally because we want to move our runtimes forward, and we want to provide a more Android-friendly interface. The Android Activity with its XML companion file is something we have discussed in the past. These community contributions are excellent, and they are what we're looking for with our runtimes being open source. 

But, we haven't committed to a higher-level API just because we're still defining what that is going to look like. We only just launched the Rive 2 Open Beta last week, and our runtimes are also still experimental.
We're going to keep this open to periodically review it to find the right way to ship it!

MightySeal:
@umberto-sonnino thanks for having a look!
Well, as an android developer, I can say that xml is #1 option to define the layout so far, that's why I added it. 
There's also [Compose](https://developer.android.com/jetpack/compose), which will make ui definition similar to elm/react/flutter, but it is only coming to beta soon and it will take time to refactor current xml layouts to compose. 

However, imo the most important thing is drawable implementation for 2 reasons:
1. It integrates with android rendering system
2. It can be easily used with compose, there will be a relatively simple wrapper for that.

I'm happy to share my knowledge, so feel free to ask/use :)

PR #87, opened 2021-04-26
Single drawable

mjohnsullivan:
Experimenting with RiveDrawable creation; notes inline

### DIFF
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
@@ -35,12 +35,6 @@ class RiveAnimationView(context: Context, attrs: AttributeSet? = null) : View(co
 
     private var resourceId: Int? = null
 
-//    var drawable: RiveDrawable
-//        get() = _drawable
-//        private set(value) {
-//            _drawable = value
-//        }
-
     var fit: Fit
         get() = drawable.fit
         set(value) {
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
@@ -370,6 +370,7 @@ class RiveAnimationView(context: Context, attrs: AttributeSet? = null) : View(co
         alignment: Alignment,
         loop: Loop,
     ) {
+        drawable.clear()
         drawable.fit = fit
         drawable.alignment = alignment
         drawable.loop = loop
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
@@ -30,15 +30,16 @@ import java.util.*
  */
 class RiveAnimationView(context: Context, attrs: AttributeSet? = null) : View(context, attrs),
     Observable<RiveDrawable.Listener> {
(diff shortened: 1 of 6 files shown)


kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -30,15 +30,16 @@ import java.util.*
...
-    private var _drawable: RiveDrawable = RiveDrawable();
-    private var resourceId: Int? = null;
-
-    var drawable: RiveDrawable
-        get() = _drawable
-        private set(value) {
-            _drawable = value
-        }
+    // There's always just one drawable associated with an animation view
+    val drawable = RiveDrawable()
```
mjohnsullivan, line 34: So here I've tried to re-use the drawable that's created when the View is instantiated

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -30,15 +30,16 @@ import java.util.*
...
+
+    private var resourceId: Int? = null
+
+//    var drawable: RiveDrawable
```
mjohnsullivan, line 38: No need for getter now that its immutable

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -316,38 +317,65 @@ class RiveAnimationView(context: Context, attrs: AttributeSet? = null) : View(co
...
+        artboardName: String? = null,
+        animationName: String? = null,
         autoplay: Boolean = drawable.autoplay,
-        fit: Fit = drawable.fit,
-        alignment: Alignment = drawable.alignment,
-        loop: Loop = drawable.loop,
+        fit: Fit = Fit.CONTAIN,
+        alignment: Alignment = Alignment.CENTER,
+        loop: Loop = Loop.NONE,
     ) {
```
mjohnsullivan, line 339: Not sure the way these initialize make sense; if we're setting the Rive file multiple times, it'll reuse the first one's artboard and animation names if they're not specified, which will probably cause new files to explode. Better to have those reset to null so defaults can be chosen?

Same wit the other parameters; best to have everything reset to defaults?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -295,17 +296,17 @@ class RiveAnimationView(context: Context, attrs: AttributeSet? = null) : View(co
...
         resourceId = resId
-        val file = File(resources.openRawResource(resId).readBytes())
-        setRiveFile(
-            file,
+        val bytes = resources.openRawResource(resId).readBytes()
+        setRiveBytes(
+            bytes,
             fit = fit,
```
mjohnsullivan, line 310: So I've moved resetting the drawable's defaults to this level; see comments on the ```setRiveFile``` function. As we eat the defaults the first time we call setRiveXXX, I just copied the defaults out to here and ```setRiveBytes```. We could have the defaults as class immutables on RiveDrawable if we wanted to keep things nicely encapsulated.

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
mjtalbot: yeah i think so

mjtalbot (approved):
LGTM

PR #88, opened 2021-04-27
Chore/gh actions

adamTrz:
Build Rive gradle lib and publish it to Maven via GH Actions. 

Included Action will trigger a build process on each release of the repo.


.github/workflows/build.yml
```diff
@@ -0,0 +1,66 @@
...
+    types: [released]
+
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.RIVE_REPO_PAT }}
```
adamTrz, line 14: This needs to be added to repos secrets.

.github/workflows/build.yml
```diff
@@ -0,0 +1,66 @@
...
+        with:
+          api-level: 29
+          script: ./gradlew publishAllPublicationsToSonatypeRepository --max-workers 1 closeAndReleaseSonatypeStagingRepository
+        env:
+          OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+          OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
+          SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
+          SIGNING_PASSWORD: ${{ secrets.SIGNING_PASSWORD }}
+          SIGNING_SECRET_KEY_RING_FILE: ${{ secrets.SIGNING_SECRET_KEY_RING_FILE }}
+          SONATYPE_STAGING_PROFILE_ID: ${{ secrets.SONATYPE_STAGING_PROFILE_ID }}
```
adamTrz, line 66: Ditto.

mjtalbot: pretty sure we got this set

app/build.gradle
```diff
@@ -30,7 +30,7 @@ dependencies {
     implementation 'androidx.core:core-ktx:1.3.1'
     implementation 'androidx.appcompat:appcompat:1.2.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.0.1'
-    implementation project(path: ':kotlin')
+    implementation 'io.github.callstack-internal:rive:0.0.5'
```
mjtalbot, line 33: guessing this is part of testing building to rive.

build.gradle
```diff
@@ -1,4 +1,16 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
+group = "com.example.library"
+version = "1.0.0"
+apply plugin: "io.github.gradle-nexus.publish-plugin"
+apply plugin: 'com.palantir.git-version'
```
mjtalbot, line 5: we've got these in kotlin/build.gradle  

(i'm assuming some of the things here are leftovers of build testing, and the main piece is the `build.yaml`?)

app/build.gradle
adamTrz: yup

build.gradle
adamTrz: Shouldn't this be in a main `build.gradle` file only?

mjtalbot: oh i commented on a dumb line. i was talking about group/version

app/build.gradle
```diff
@@ -38,7 +38,7 @@ dependencies {
     implementation 'androidx.core:core-ktx:1.3.2'
     implementation 'androidx.appcompat:appcompat:1.2.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.0.4'
-    implementation project(path: ':kotlin')
+    implementation 'app.rive:rive-android:0.0.3'
```
mjtalbot, line 41: (i'm not sure what best practice is) 

does it make sense for the sample app to point at a published version, or to use our local kotlin changes?

build.gradle
```diff
@@ -1,4 +1,16 @@
...
+
+def details = versionDetails()
+
+ext {
```
mjtalbot, line 9: this (probably) conflicts with the ext block in kotlin/build.gradle

```
ext {
    PUBLISH_GROUP_ID = 'app.rive'
    PUBLISH_VERSION = '0.0.3'
    PUBLISH_ARTIFACT_ID = 'rive-android'
}
```

mjtalbot: should this block update that block? (also i'm not sure if we should be going for rive or rive-android. I guess android is pretty much implied here.  i think we're probably going to fuss over it for a little bit until we make an announcement :D)

mjtalbot: or just remove this block i dont think it makes a difference which file has this setup. we're only interested in pushing one after all :D

app/build.gradle
adamTrz: Dunno tbh, just did it to test if it actually works ;)

mjtalbot: :D

build.gradle
adamTrz: Yeah, but we need to bump version, otherwise Maven will get angry - we can't release sam version.

mjtalbot: should we keep PUBLISH_ARTIFACT_ID as rive-android for now? (just because that's what we've released under so far)

mjtalbot (approved):
lgtm. i'm still fussing over the artifact_id but honestly i think im just fussing

PR #118, opened 2021-06-02
Adds high level constructor for Rive files over http

mjohnsullivan:
Also updates some dependency versions.

Volley is now a dependency.

### DIFF
--- a/kotlin/src/main/AndroidManifest.xml
+++ b/kotlin/src/main/AndroidManifest.xml
@@ -2,6 +2,9 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
+    <uses-permission android:name="android.permission.INTERNET" />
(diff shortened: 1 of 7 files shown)


kotlin/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,9 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
+    <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
mjtalbot, line 6: what does this mean for our library. does it mena it now requires internet access? or it just kinda says that it wants it, but you can still not give it if you don't want to ?

mjtalbot (approved):
lgtm. i feel some version fun coming on but thats about it :P

PR #135, opened 2021-06-22
Throw specific rive exceptions

avivian:
We now throw the following errors: 
```
RiveException
MalformedFileException
UnsupportedRuntimeVersionException
ArtboardException
AnimationException
StateMachineException
StateMachineInputException
```


cpp/src/helpers/general.cpp
```diff
@@ -146,15 +146,15 @@ namespace rive_android
 		}
 		else if (result == rive::ImportResult::unsupportedVersion)
 		{
-			return throwRiveError("Unsupported Rive File Version.");
+			return throwUnsupportedRuntimeVersionException("Unsupported Rive File Version.");
```
mjtalbot, line 149: need to rebuild cpp for this change. (i've not checked if you have yet :P)

avivian: It's totes done

mjtalbot (approved):
lgtm. you may as well do a build with this 👍

PR #149, opened 2021-07-23
Deploy on merge

mjtalbot:
Brings android in line with ios for doing releases. 

we should line up how we do changelogs, I've really not touched it here or on ios. + kinda fine with it just being whatever goes into commits/pr's tbh. sure maybe we want special controls for ignoring things, but i think we can overthink all of this too. i guess grouping features would look nice though eventually. 

I know you had some thoughts on that @luigi-rosso.


.github/scripts/release/.release-it.json
```diff
@@ -0,0 +1,34 @@
...
+      },
+      "out": {
+        "file": "../../../VERSION",
+        "type": "text/plain"
+      }
+    }
+  },
+  "hooks": {
+    "after:bump": [
+      "npx auto-changelog -p -o ../../../CHANGELOG.md --hide-empty-releases",
```
luigi-rosso, line 29: More of a question, but how come changelog is defined twice (in the git section and here in after:bump). Can we remove the one from the git section above? I was curious about this when I was doing the dart one too and wasn't sure.

luigi-rosso (approved):
LGTM, just a question.

.github/scripts/release/.release-it.json
mjtalbot: hah yeah what is that about. ill take it out and see what the dry runs say. it doesnt look like it should do anything really. the 2nd one is the important one

mjtalbot: oh i think its this 

Changelog
By default, release-it generates a changelog, to show and help select a version for the new release. Additionally, this changelog serves as the release notes for the GitHub or GitLab release.

The default command is based on git log .... This setting (git.changelog) can be overridden. To further customize the release notes for the GitHub or GitLab release, there's github.releaseNotes or gitlab.releaseNotes. Make sure any of these commands output the changelog to stdout. Plugins are available for:

GitHub and GitLab Releases
auto-changelog
Conventional Changelog
Keep A Changelog
→ See Changelog for more details.

mjtalbot: its for the github release notes i guess.... still pretty unsure about the use of those in our context

PR #154, opened 2021-08-19
Track playing animations on detach from window

avivian:
Fixes an issue where the animation no longer plays when reattaching to the window.

Not super sure how to test this. I think there must be a way to trigger the onDetach and onAttach callbacks, but struggling to work it out. Putting this up for feedback in the mean time. Tested a similar implementation via the react native runtime however and it's fixed the issue.

### DIFF
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
@@ -48,6 +48,8 @@ class RiveAnimationView(context: Context, attrs: AttributeSet? = null) : View(co
 
     private var resourceId: Int? = null
 
+    private var _detachedState: DetachedRiveState? = null;
+
     var fit: Fit
         get() = drawable.fit
         set(value) {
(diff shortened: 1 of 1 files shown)


kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -446,12 +448,29 @@ class RiveAnimationView(context: Context, attrs: AttributeSet? = null) : View(co
         requestLayout()
     }
 
-
     override fun onDetachedFromWindow() {
         super.onDetachedFromWindow()
+        // Track the playing animations and state machines so we can resume them if the window is
+        // attached.
+        _detachedState = DetachedRiveState(
+            playingAnimationsNames = playingAnimations.map { it.animation.name },
```
mjtalbot, line 456: looks good i think!

this'll only do the "natural" play on animations, so i think you may need to track loop mode etc as well, as it'll overwrite that otherwise (if you specified a different loop mode when initially hitting play) but i think we can probably add that later too... this feels like it gets us pretty far.

mjtalbot (approved):
lgtm

PR #171, opened 2021-11-10
update to use relinker

mjtalbot:
we can also go for versioned .so files

i think that means that we will want to supply those though as part of building our app (which we can do)

```
Versioning
In the event that your library's code is changed, it is a good idea to specify a specific version. Doing so will allow ReLinker to update the workaround library file successfully. In the case that the system handles the library loading appropriately, the version specified is not used as all library files are extracted and replaced on update or install.

To specify a version for your library simply provide it as an additional parameter for loadLibrary like:

ReLinker.loadLibrary(context, "mylibrary", "1.0");
This will cause ReLinker to look for, and load libmylibrary.so.1.0. Subsequent version updates will automatically clean up all other library versions.
```

we could track a version number in Rive.kt  and update it when we build the .so's with something like this  (im sure someone with more bash foo can make that nicer)

```
# https://stackoverflow.com/questions/59895/how-can-i-get-the-source-directory-of-a-bash-script-from-within-the-script-itsel
SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"

# increment the SOVersion by 1. 
RIVE_KT=$SCRIPT_DIR/../kotlin/src/main/java/app/rive/runtime/kotlin/core/Rive.kt
CURRENT_VERSION=`cat $RIVE_KT |grep "SOVersion = " | grep -Eo "\d+"`
NEW_VERSION=$((CURRENT_VERSION+1))

sed -i '' 's/SOVersion = "'"$CURRENT_VERSION"'"/SOVersion = "'"$NEW_VERSION"'"/' $RIVE_KT
```

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -44,6 +44,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     implementation 'androidx.core:core-ktx:1.5.0'
     implementation 'androidx.appcompat:appcompat:1.3.0'
+    implementation 'com.getkeepsafe.relinker:relinker:1.4.4'
     implementation 'androidx.startup:startup-runtime:1.1.0'
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'androidx.test:core-ktx:1.3.0'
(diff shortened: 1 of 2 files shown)


umberto-sonnino (approved):
lgtm!

PR #174, opened 2021-11-10
Swappy callbacks v2

umberto-sonnino:
- Callbacks & drawing are now working from cpp >> kotlin
- Cleaned up the old OpenGL Activity since we're using the Surface View now
- Using now cached `jclass` and `jmethodID`s so we don't need to re-fetch them in the `draw()` callback


kotlin/src/main/java/app/rive/runtime/kotlin/core/Artboard.kt
```diff
@@ -173,12 +170,7 @@ class Artboard(val cppPointer: Long) {
      * [elapsedTime] is currently not taken into account.
      */
     fun advance(elapsedTime: Float): Boolean {
-        var didUpdate = false
-        playableInstances.forEach { instance ->
-            instance.apply(this, elapsedTime)
-        }
-
-        return cppAdvance(cppPointer, elapsedTime) || didUpdate
+        return cppAdvance(cppPointer, elapsedTime)
```
mjtalbot, line 173: not really sure what this does/did

kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererSkia.kt
```diff
@@ -45,8 +41,37 @@ class RendererSkia : BaseRenderer() {
         startFrame(cppPointer)
     }
 
+    fun play(animationName: String) {
```
mjtalbot, line 44: so are you putting the render controls into the rendererSkia piece? as opposed to the view?

mjtalbot (approved):
lets get this in.. just a few q's but nothing to stop merging.

kotlin/src/main/java/app/rive/runtime/kotlin/core/Artboard.kt
umberto-sonnino: `advance()` can return a bool to signal that is still advancing
I added this during my explorations, I noticed that it wasn't yet supported! 
We can use this to infer stuff like when a OneShot animation has been completed

kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererSkia.kt
umberto-sonnino: I put it here for now just to get it rolling, but I was actually thinking of pulling all this out in its own object just to separate responsibilities. What do you think?

mjtalbot: yeah sounds good! (that could essentially be a renamed version of what riveDrawable is right now, without drawable/animatable) 

i still think we'll want a base view class that doesnt use that object at all, and just deals with the handover between kotlin/cpp.

PR #175, opened 2021-11-11
Custom fit & alignment, `RivePlayer` class

umberto-sonnino:
- Use a separate class called `RivePlayer` for managing the playback of the animations and `Fit`/`Alignment` for the active artboard
- Pass `Fit` & `Alignment` via XML


kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererSkia.kt
```diff
@@ -24,59 +20,105 @@ class RendererSkia : BaseRenderer() {
...
+//        startFrame(cppPointer)
         cppDraw(artboard.cppPointer, cppPointer)
 //        var start = SystemClock.elapsedRealtimeNanos()
 //        artboard.drawSkia(this)
 //        val now = SystemClock.elapsedRealtimeNanos()
 //        Log.d("SKIA DRAW", "Frame: ${(now - start) / 1000000} ms")
     }
+
+
+    private class RivePlayer {
```
mjtalbot, line 71: this is the thing that'll get the rest of drawable dumped ontop of i guess?

PR #177, opened 2021-11-18
Miscellaneous fixes

umberto-sonnino:
- Perform UI updates on UI thread
- Fix the cleanup step so the Renderer gets deallocated
- Make sure we queue another callback when we're running
- Let `RiveAnimationView` initialize Swappy (and comment out `RenderMetrics` for now)
- Few other minor things

### DIFF
--- a/app/src/main/java/app/rive/runtime/example/AndroidPlayerActivity.kt
+++ b/app/src/main/java/app/rive/runtime/example/AndroidPlayerActivity.kt
@@ -337,12 +337,14 @@ class AndroidPlayerActivity : AppCompatActivity() {
                     text = animation.stateMachine.name
                 }
                 text?.let { theText ->
-                    val textView = TextView(that)
-                    textView.text = "Play $theText"
-                    events.addView(textView, 0)
-                    playButtonMap[theText]?.background?.setTint(Color.GREEN)
-                    pauseButtonMap[theText]?.background?.setTint(Color.WHITE)
-                    stopButtonMap[theText]?.background?.setTint(Color.WHITE)
+                    runOnUiThread {
+                        val textView = TextView(that)
+                        textView.text = "Play $theText"
+                        events.addView(textView, 0)
+                        playButtonMap[theText]?.background?.setTint(Color.GREEN)
+                        pauseButtonMap[theText]?.background?.setTint(Color.WHITE)
+                        stopButtonMap[theText]?.background?.setTint(Color.WHITE)
+                    }
                 }
             }
 
@@ -354,12 +356,14 @@ class AndroidPlayerActivity : AppCompatActivity() {
                     text = animation.stateMachine.name
                 }
                 text?.let {
-                    val textView = TextView(that)
-                    textView.text = "Pause $text"
-                    events.addView(textView, 0)
-                    playButtonMap[text]?.background?.setTint(Color.WHITE)
-                    pauseButtonMap[text]?.background?.setTint(Color.BLUE)
-                    stopButtonMap[text]?.background?.setTint(Color.WHITE)
+                    runOnUiThread {
+                        val textView = TextView(that)
+                        textView.text = "Pause $text"
+                        events.addView(textView, 0)
+                        playButtonMap[text]?.background?.setTint(Color.WHITE)
+                        pauseButtonMap[text]?.background?.setTint(Color.BLUE)
+                        stopButtonMap[text]?.background?.setTint(Color.WHITE)
+                    }
                 }
             }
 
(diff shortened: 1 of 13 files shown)


app/src/main/java/app/rive/runtime/example/AndroidPlayerActivity.kt
```diff
@@ -337,12 +337,14 @@ class AndroidPlayerActivity : AppCompatActivity() {
                     text = animation.stateMachine.name
                 }
                 text?.let { theText ->
-                    val textView = TextView(that)
-                    textView.text = "Play $theText"
-                    events.addView(textView, 0)
-                    playButtonMap[theText]?.background?.setTint(Color.GREEN)
-                    pauseButtonMap[theText]?.background?.setTint(Color.WHITE)
-                    stopButtonMap[theText]?.background?.setTint(Color.WHITE)
+                    runOnUiThread {
```
mjtalbot, line 340: oh. whats going on here? is this "just" an improvement or does this fix a bug that was before the swappy change?

cpp/src/bindings/bindings_animation_view.cpp
```diff
@@ -93,22 +82,19 @@ extern "C"
...
+  Java_app_rive_runtime_kotlin_RiveAnimationView_cppInit(
+      JNIEnv *env, jobject view,
+      jobject activity, jlong initialSwapIntervalNS)
   {
-    ANativeWindow *surfaceWindow = ANativeWindow_fromSurface(env, surface);
-    auto skiaRenderer = (JNIRendererSkia *)rendererAddr;
-    skiaRenderer->setWindow(surfaceWindow);
-  }
+    // Should never happen
+    if (Swappy_version() != SWAPPY_PACKED_VERSION)
```
mjtalbot, line 90: 👀  how would this happen? (i'm guessing the check is there because this has happened?

kotlin/src/androidTest/java/app/rive/runtime/kotlin/core/RiveAnimationConfigurationsTest.kt
```diff
@@ -79,9 +79,9 @@ class RiveAnimationConfigurationsTest {
     @Test
     fun checkdurations1sec120fps_f30f50() {
         val appContext = initTests()
-        var file =
+        val file =
             File(appContext.resources.openRawResource(R.raw.animationconfigurations).readBytes())
-        var animation = file.firstArtboard.animation("1sec120fps_f50f80")
+        val animation = file.firstArtboard.animation("1sec120fps_f50f80")
```
mjtalbot, line 84: 👍

app/src/main/java/app/rive/runtime/example/AndroidPlayerActivity.kt
umberto-sonnino: This was crashing with the Swappy update because these `notifyPlay/Pause/etc.()` are now triggered from the cpp render thread.
Android doesn't let it happen though and it crashes, hence the need to force it to happen on the UI thread

cpp/src/bindings/bindings_animation_view.cpp
umberto-sonnino: This is copypasta from Google directly. It never happened for me, and I can remove it if you think this is just confusing

kotlin/src/androidTest/java/app/rive/runtime/kotlin/core/RiveViewTest.kt
```diff
@@ -452,7 +452,7 @@ class RiveViewTest {
             view.setRiveResource(R.raw.multiple_animations, autoplay = false)
 
             view.play("one", Loop.PINGPONG)
-            view.drawable.advance(100f)
+            view.drawable.advance(0.1f)
```
mjtalbot, line 455: i think this was testing some specific behaviour with how advancing past  the end of an animation works, but i'm really not sure. 

i think this change to the test still makes sense regardless (as we're meant to be testing stopping/pausing animations) 

But we should make a note and add some "over-advance" tests to make sure we test how advancing past the end of things works properly

umberto-sonnino: Yeah, there's a few of these that are still failing that I need to look into. 
I made this one change because we're advancing in `ms` now instead of `ns`, which we previously converted dividing by `1000`

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -215,7 +230,9 @@ class RiveAnimationView(context: Context, attrs: AttributeSet? = null) :
 //                1e3f / fps
 //            )
 //        Trace.endSection()
-        Choreographer.getInstance().postFrameCallback(this)
+        if (isRunning) {
```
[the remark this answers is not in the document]
umberto-sonnino: Unless we add `this` as a callback, `doFrame()` won't get called..!

kotlin/src/main/java/app/rive/runtime/kotlin/core/File.kt
```diff
@@ -58,13 +58,9 @@ class File(bytes: ByteArray) {
...
         if (artboardPointer == 0L) {
             throw ArtboardException(
-                "Artboard \"$name\" not found.\nAvailable Artboards:\n${
-                    artboardNames.forEach {
-                        println(
-                            "- \"$it\""
-                        )
-                    }
-                }"
+                "Artboard \"$name\" not found. Available Artboards: ${artboardNames.map {
```
mjtalbot, line 61: :psyduck:

kotlin/src/main/java/app/rive/runtime/kotlin/renderers/BaseRenderer.kt
```diff
@@ -1,15 +1,9 @@
 package app.rive.runtime.kotlin.renderers
 
+import android.util.Log
```
mjtalbot, line 3: probably leftover, but also not important

kotlin/src/main/java/app/rive/runtime/kotlin/renderers/BaseRenderer.kt
```diff
@@ -1,15 +1,9 @@
...
+
 abstract class BaseRenderer {
     internal abstract var cppPointer: Long
 
     protected abstract fun cleanupJNI(cppPointer: Long)
```
mjtalbot, line 8: probably worth making a note to see about some cleanup things as part of tests (we should have a "load" test for loading in more and more animations at the same time... or something. but also not the most crucial)

mjtalbot (approved):
lgtm! just comments as i'm trying to stay somewhat in touch with changes :D

kotlin/src/main/java/app/rive/runtime/kotlin/core/File.kt
umberto-sonnino: Hope this helps with those "empty spaces at the end" problems

app/src/main/java/app/rive/runtime/example/AndroidPlayerActivity.kt
mjtalbot: oh interesting. for later: i wonder if we can wrap the place where we call notifyPlay in our "event handler thing" with runOnUiThread {} so the user doesn't have to think.

umberto-sonnino: yeah, good point, we probably should

kotlin/src/main/java/app/rive/runtime/kotlin/core/File.kt
mjtalbot: yeah this side looks much nicer, the previous looks very wtf

cpp/src/bindings/bindings_animation_view.cpp
mjtalbot: dunno. maybe just link the pasta sauce, so if there's something with it whoever is looking at it has some context as to why its here. or don't.  i guess its fine :D

PR #178, opened 2021-11-18
Swappy measure

umberto-sonnino:
- Fix `onMeasure()`for surface view
- Add `DEBUG` flag that removes `LOG`s and logging `thread`


mjtalbot (approved):
lgtm. -3/4kb on .so's too 🙌

PR #179, opened 2021-11-18
Swappy tests

umberto-sonnino:
Refactored the tests setup to include a `MockRenderer` that sets up the most basic cpp structures.
Make sure we `advance()` upon setting Rive resource and when firing states in tests.
In fact, this is slightly different now having a thread running in the background and managing playback: in fact, in our mock, we don't yet instantiate the OpenGL surface and all the jazz that comes along with it so, when we need to validate that a state has fired, we have to manually advance it.

### DIFF
--- a/kotlin/src/androidTest/java/app/rive/runtime/kotlin/core/RiveEventTest.kt
+++ b/kotlin/src/androidTest/java/app/rive/runtime/kotlin/core/RiveEventTest.kt
@@ -17,22 +17,24 @@ class Observer : RiveDrawable.Listener {
     var stops = mutableListOf<PlayableInstance>()
     var loops = mutableListOf<PlayableInstance>()
     var states = mutableListOf<StateChanged>()
-    override fun notifyPlay(playableInstance: PlayableInstance) {
-        plays.add(playableInstance)
+
+    override fun notifyPlay(animation: PlayableInstance) {
+        plays.add(animation)
     }
 
-    override fun notifyPause(playableInstance: PlayableInstance) {
-        pauses.add(playableInstance)
+    override fun notifyPause(animation: PlayableInstance) {
+        pauses.add(animation)
     }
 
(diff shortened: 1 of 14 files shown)


kotlin/src/androidTest/java/app/rive/runtime/kotlin/core/RiveEventTest.kt
```diff
@@ -174,7 +169,7 @@ class RiveEventTest {
             view.play("one", Loop.ONESHOT)
 
             view.drawable.advance(
-                view.drawable.animations.first().animation.effectiveDurationInSeconds * 1000 + 1
+                view.drawable.animations.first().animation.effectiveDurationInSeconds + 1
```
mjtalbot, line 172: this is part of the milli to nano change?

kotlin/src/androidTest/java/app/rive/runtime/kotlin/core/RiveEventTest.kt
```diff
@@ -336,13 +329,14 @@ class RiveEventTest {
...
     fun viewBlendState1DBroken() {
         UiThreadStatement.runOnUiThread {
-            val appContext = initTests()
             val observer = Observer()
 
             val view = RiveAnimationView(appContext)
             view.registerListener(observer)
             view.setRiveResource(R.raw.blend_state, stateMachineName = "one")
             view.fireState("one", "blend mix")
+            view.drawable.advance(0.0f)
```
mjtalbot, line 338: 👀  i guess before advance was called somewhere else?

mjtalbot (approved):
lgtm

there are a few extra advance calls after state machine triggers that make me think we may have changed behaviour here slightly

but not sure if it makes a difference for the runtime tbh. 
same with ms vs ns. i dont think its bad to change. but i'm unsure if it makes a difference for our users. i suspect not?

PR #183, opened 2021-11-23
Run clang-format on all

umberto-sonnino:
Just realized we didn't have the proper formatter set up in this repo, I figured it was worth a quick fix

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,37 @@
+---
+BasedOnStyle: LLVM
+IndentWidth: 4
+UseTab: ForIndentation
+TabWidth: 4
+BreakBeforeBraces: Allman
+IndentCaseLabels: true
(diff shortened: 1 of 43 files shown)


umberto-sonnino:
> lgtm. would be cool to put setup instructions for this into readme (im guessing there's a clang format thing in vscode too? going to look...)

Yeah, this actually ties nicely with VSCode so whenever you autoformat, it uses the settings in that `.clang-format` file I added.
I just ran it on our C++ folders like this:
```
cpp % find ./src -type f -name '*pp' | xargs clang-format -i
cpp % find ./include -type f -name '*.h' | xargs clang-format -i 
```
But we shouldn't need to run it manually every time.

We could set a pre-commit hook that does it for us, though.

PR #198, opened 2021-11-26
190 add optional tracing

### LINKED ISSUE #190: Allowing tracing as part of JNI Renderer Skia

mjtalbot:
fixes #190 

optional add tracing. maybe worth double checking if there's a better way to make sure when its not enabled it doesn't eat up resources, I'm sure @luigi-rosso might have ideas on this too. 

we could also compile this in/out if we wanted, it would be nice to be able to enable/disable this on demand though

### DIFF
--- a/app/src/main/java/app/rive/runtime/example/MetricsActivity.kt
+++ b/app/src/main/java/app/rive/runtime/example/MetricsActivity.kt
@@ -11,18 +11,32 @@ class MetricsActivity : AppCompatActivity() {
     private val containerView by lazy(LazyThreadSafetyMode.NONE) {
         findViewById<LinearLayout>(R.id.container)
     }
+//
+//    private fun addSwappyView() {
+//        findViewById(R.id.swappy_view)
+//            ?: RiveAnimationView(this, null).also {
+//                val density = resources.displayMetrics.density
+//                it.layoutParams = LinearLayout.LayoutParams(
+//                    LinearLayout.LayoutParams.MATCH_PARENT,
+//                    (366 * density).toInt(),
+//                )
+//                containerView.addView(it, 0)
+//
+//            }
+//    }
 
-    private fun addSwappyView() {
-        findViewById(R.id.swappy_view)
-            ?: RiveAnimationView(this, null).also {
-                val density = resources.displayMetrics.density
-                it.layoutParams = LinearLayout.LayoutParams(
-                    LinearLayout.LayoutParams.MATCH_PARENT,
-                    (366 * density).toInt(),
-                )
-                containerView.addView(it, 0)
-            }
-    }
+//    private fun updateFps() {
+//
+//        val fpsView = findViewById<TextView>(R.id.fps)
+//        val fps = nGetAverageFps(riveRenderer.address)
+//        fpsView?.text =
+//            java.lang.String.format(
+//                Locale.US,
+//                "Frame rate: %.1f Hz (%.2f ms)",
+//                fps,
+//                1e3f / fps
+//            )
+//    }
 
 
     override fun onCreate(savedInstanceState: Bundle?) {
@@ -31,6 +45,6 @@ class MetricsActivity : AppCompatActivity() {
         Rive.init(this)
         setContentView(R.layout.activity_metrics)
         // Make sure this is initialized.
-        addSwappyView()
+//        addSwappyView()
     }
 }
--- a/app/src/main/java/app/rive/runtime/example/MetricsActivity.kt
+++ b/app/src/main/java/app/rive/runtime/example/MetricsActivity.kt
@@ -3,7 +3,6 @@ package app.rive.runtime.example
 import android.os.Bundle
 import android.widget.LinearLayout
 import androidx.appcompat.app.AppCompatActivity
-import app.rive.runtime.kotlin.RiveAnimationView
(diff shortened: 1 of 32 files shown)


kotlin/src/main/res/values/attrs.xml
```diff
@@ -2,6 +2,7 @@
 <resources>
     <declare-styleable name="CustomRiveView">
         <attr name="riveAutoPlay" format="boolean" />
+        <attr name="riveTraceAnimations" format="boolean" />
```
umberto-sonnino, line 5: Oh, really cool that we can define it also in XML!

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -142,11 +142,14 @@ open class RiveAnimationView(context: Context, attrs: AttributeSet? = null) :
...
                 val autoplay =
                     getBoolean(R.styleable.RiveAnimationView_riveAutoPlay, defaultAutoplay)
+                val riveTraceAnimations =
+                    getBoolean(R.styleable.RiveAnimationView_riveTraceAnimations, false)
                 val artboardName = getString(R.styleable.RiveAnimationView_riveArtboard)
                 val animationName = getString(R.styleable.RiveAnimationView_riveAnimation)
                 val stateMachineName = getString(R.styleable.RiveAnimationView_riveStateMachine)
                 val resourceId = getResourceId(R.styleable.RiveAnimationView_riveResource, -1)
                 val url = getString(R.styleable.RiveAnimationView_riveUrl)
+                renderer = RiveArtboardRenderer(autoplay = autoplay, trace=riveTraceAnimations)
```
umberto-sonnino, line 152: Kotlin formatter also might want a pass here. I think it's `CMD + Alt + L` in Android Studio

app/src/main/java/app/rive/runtime/example/MetricsActivity.kt
```diff
@@ -11,18 +11,32 @@ class MetricsActivity : AppCompatActivity() {
     private val containerView by lazy(LazyThreadSafetyMode.NONE) {
         findViewById<LinearLayout>(R.id.container)
     }
+//
```
umberto-sonnino, line 14: Cleanup! Any reason to leave it hanging around? Otherwise, we can just delete it

cpp/include/models/jni_renderer_skia.hpp
```diff
@@ -51,34 +55,27 @@ namespace rive_android
...
 
 		jobject mKtRenderer;
 
 		SkCanvas* mGpuCanvas;
 		rive::SkiaRenderer* mSkRenderer;
+		ITracer* mTracer;
 
 	public:
-		JNIRendererSkia(jobject ktObject) :
+		JNIRendererSkia(jobject ktObject, bool trace=false) :
```
umberto-sonnino, line 66: I added `.clang-format` in the project root folder, you might be able to just run the formatter via VSCode to fix some minor styles here.

cpp/include/models/jni_renderer_skia.hpp
```diff
@@ -51,34 +55,27 @@ namespace rive_android
...
-			if (lib != NULL)
-			{
-				//  Retrieve function pointers from shared object.
-				ATrace_beginSection = reinterpret_cast<fp_ATrace_beginSection>(
-				    dlsym(lib, "ATrace_beginSection"));
-				ATrace_endSection = reinterpret_cast<fp_ATrace_endSection>(
-				    dlsym(lib, "ATrace_endSection"));
-				ATrace_isEnabled = reinterpret_cast<fp_ATrace_isEnabled>(
-				    dlsym(lib, "ATrace_isEnabled"));
+			bool traceAvailable = android_get_device_api_level()>=23;
```
umberto-sonnino, line 69: Oh that's cool that you can check that natively

umberto-sonnino (approved):
LGTM! 
We should really find a way to run our formatters as a pre-commit hook.

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
mjtalbot: 🙈

app/src/main/java/app/rive/runtime/example/MetricsActivity.kt
mjtalbot: oh left it in just because i figured i'd take a proper look once the choreographer is in (and i moved some fps thing from the rive view in here )

cpp/include/models/jni_renderer_skia.hpp
mjtalbot: oh yeah how do i run this!

cpp/include/models/jni_renderer_skia.hpp
mjtalbot: yeah right. its super neat!

cpp/include/models/jni_renderer_skia.hpp
umberto-sonnino: The shortcut is `Alt+Shift+F`, there's also the option in the command palette to `Format Document`

PR #202, opened 2021-12-02
Mock Choreographer

umberto-sonnino:


app/src/main/java/app/rive/runtime/example/LowLevelActivity.kt
```diff
@@ -52,11 +52,6 @@ class LowLevelRiveView(context: Context) : RiveTextureView(context) {
 
     private var bounds: AABB = AABB(100f, 100f)
 
-    override fun onAttachedToWindow() {
```
mjtalbot, line 55: oh amazing

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -128,59 +129,83 @@ open class RiveAnimationView(context: Context, attrs: AttributeSet? = null) :
     val playingStateMachines: HashSet<StateMachineInstance>
         get() = renderer.playingStateMachines
 
-    init {
+    protected data class RendererAttrs(
+        val alignmentIndex: Int = 4,
```
mjtalbot, line 133: oh cool! (we should get these defaults ironed out. but later :P)

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -429,11 +427,10 @@ open class RiveArtboardRenderer(
                 }
 
             }
-
         } else {
             this.activeArtboard?.advance(0f)
+            start()
```
mjtalbot, line 432: 👀  we only run start if we don't have a selected artboard?

mjtalbot (approved):
lgtm

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
umberto-sonnino: The other branch of this `if` has a few `play()` calls that internally call `start()` already. This is just to prevent it being called multiple times unnecessarily!

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
umberto-sonnino: Yeah, I copypasta'd those defaults from underneath, we can extract them in const values for clarity

PR #204, opened 2021-12-02
Smallv2fixes

mjtalbot:
will rebase on https://github.com/rive-app/rive-android/pull/202 once merged. 

just addresses the things i ran into using our lib for the piggy bank app

### DIFF
--- a/kotlin/src/main/AndroidManifest.xml
+++ b/kotlin/src/main/AndroidManifest.xml
@@ -1,1 +1,17 @@
-<manifest package="app.rive.runtime.kotlin" />
+<manifest package="app.rive.runtime.kotlin"
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools">
+
+    <application>
+        <provider
+            android:name="androidx.startup.InitializationProvider"
+            android:authorities="${applicationId}.androidx-startup"
+            android:exported="false"
+            tools:node="merge">
+            <meta-data
+                android:name="app.rive.runtime.kotlin.RiveInitializer"
+                android:value="androidx.startup"
+                tools:node="remove" />
+        </provider>
+    </application>
+</manifest>
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/controllers/RiveController.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/controllers/RiveController.kt
@@ -4,11 +4,11 @@ abstract class RiveController<T> {
     var isActive = false
         protected set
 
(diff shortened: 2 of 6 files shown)


kotlin/src/main/java/app/rive/runtime/kotlin/controllers/RiveController.kt
```diff
@@ -4,11 +4,11 @@ abstract class RiveController<T> {
     var isActive = false
         protected set
 
-    open fun initialize(core: T): Boolean {
+    open fun initialize(artboard: T): Boolean {
```
umberto-sonnino, line 7: Oh, I forgot to remove these Controllers from here, going to give it a pass next

kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererMetrics.kt
```diff
@@ -26,7 +26,7 @@ class RendererMetrics(activity: Activity) : Window.OnFrameMetricsAvailableListen
         @Suppress("DEPRECATION")
         val display = wm.defaultDisplay
         val refreshRateHz = display.refreshRate
-        val refreshPeriodNanos = (ONE_S_IN_NS / refreshRateHz).toLong()
+//        val refreshPeriodNanos = (ONE_S_IN_NS / refreshRateHz).toLong()
```
umberto-sonnino, line 29: Was this causing issues?

kotlin/src/main/AndroidManifest.xml
```diff
@@ -1 +1,17 @@
-<manifest package="app.rive.runtime.kotlin" />
\ No newline at end of file
+<manifest package="app.rive.runtime.kotlin"
```
umberto-sonnino, line 1: Oh! Initializer's back in the manifest! Does this mean we no longer have to include it in the example app Manifest?

kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererMetrics.kt
mjtalbot: oh no idea. just a couple of things that werent used so i commented them out for the time being. figure his renderermetrics.kt things isnt quite used yet?

kotlin/src/main/AndroidManifest.xml
mjtalbot: no we still do :P i think, as its in, but disabled, its really just in to suppress the warning i get building our app. we can enable it though, and ask our users to either disable it or not

umberto-sonnino (approved):

PR #205, opened 2021-12-02
tore the drawSkia piece apart a bit

mjtalbot:
what i think we need to do with the drawSkia piece

*probably dont merge, more for talking point*

### DIFF
--- a/app/src/main/java/app/rive/runtime/example/LowLevelActivity.kt
+++ b/app/src/main/java/app/rive/runtime/example/LowLevelActivity.kt
@@ -33,7 +33,14 @@ class LowLevelRiveView(context: Context) : RiveTextureView(context) {
     // Initialize renderer first: we can't create Files without one.
     override val renderer = object : RendererSkia() {
         override fun draw() {
-            artboard.drawSkia(cppPointer, Fit.COVER, Alignment.CENTER)
+            artboard?.let {
+                save()
+                align(Fit.COVER, Alignment.CENTER,  AABB(width,height), it.bounds)
+                it.drawSkia(
+                    cppPointer
+                )
+                restore()
+            }
         }
 
         override fun advance(elapsed: Float) {
--- a/cpp/include/bindings/bindings_renderer.hpp
+++ b/cpp/include/bindings/bindings_renderer.hpp
@@ -100,6 +100,22 @@ extern "C"
 	JNIEXPORT void JNICALL
 	Java_app_rive_runtime_kotlin_renderers_RendererSkia_cppClearSurface(
 	    JNIEnv*, jobject, jlong rendererRef);
+
+	JNIEXPORT void JNICALL
+	Java_app_rive_runtime_kotlin_renderers_RendererSkia_cppSave(
+	    JNIEnv*, jobject, jlong rendererRef);
+
+	JNIEXPORT void JNICALL
+	Java_app_rive_runtime_kotlin_renderers_RendererSkia_cppRestore(
+	    JNIEnv*, jobject, jlong rendererRef);
+
+	JNIEXPORT jint JNICALL
+	Java_app_rive_runtime_kotlin_renderers_RendererSkia_cppWidth(
+	    JNIEnv*, jobject, jlong rendererRef);
+
+	JNIEXPORT jint JNICALL
+	Java_app_rive_runtime_kotlin_renderers_RendererSkia_cppHeight(
+	    JNIEnv*, jobject, jlong rendererRef);
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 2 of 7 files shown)


cpp/include/bindings/bindings_renderer.hpp
```diff
@@ -100,6 +100,22 @@ extern "C"
...
+	JNIEXPORT void JNICALL
+	Java_app_rive_runtime_kotlin_renderers_RendererSkia_cppSave(
+	    JNIEnv*, jobject, jlong rendererRef);
+
+	JNIEXPORT void JNICALL
+	Java_app_rive_runtime_kotlin_renderers_RendererSkia_cppRestore(
+	    JNIEnv*, jobject, jlong rendererRef);
+
+	JNIEXPORT jint JNICALL
+	Java_app_rive_runtime_kotlin_renderers_RendererSkia_cppWidth(
```
mjtalbot, line 113: is it odd that the renderer knows its width and height?

kotlin/src/main/java/app/rive/runtime/kotlin/core/Artboard.kt
```diff
@@ -35,8 +35,7 @@ class Artboard(val cppPointer: Long) {
 
     // TODO: this will be a cppDraw call after we remove our old renderer.
```
mjtalbot, line 36: yeah we got a few leftovers, there's also the drawGL to consider too.. i wonder if there's a neat way to keep this complexity away from our users. probably not the biggest concern RIGHT now. but once we sort this it'll be another breaking change no doubt

umberto-sonnino: Yeah, we've gone back and forth on this, do we want to keep the old Renderer around at all? IIRC we were going to just remove it altogether in favor of Skia

app/src/main/java/app/rive/runtime/example/LowLevelActivity.kt
```diff
@@ -33,7 +33,14 @@ class LowLevelRiveView(context: Context) : RiveTextureView(context) {
...
         override fun draw() {
-            artboard.drawSkia(cppPointer, Fit.COVER, Alignment.CENTER)
+            artboard?.let {
+                save()
+                align(Fit.COVER, Alignment.CENTER,  AABB(width,height), it.bounds)
+                it.drawSkia(
+                    cppPointer
+                )
+                restore()
+            }
```
umberto-sonnino, line 43: Mmh is it ok to ask our users to do this when running your own loop? 
Maybe it is! Otherwise, we can keep the old call around in the artboard API for ease of use?

kotlin/src/main/java/app/rive/runtime/kotlin/core/Artboard.kt
```diff
@@ -187,8 +186,8 @@ class Artboard(val cppPointer: Long) {
     /**
      * Draw the the artboard to the [renderer].
      */
-    fun drawSkia(rendererAddress: Long, fit: Fit, alignment: Alignment) {
-        cppDrawSkia(cppPointer, rendererAddress, fit, alignment)
+    fun drawSkia(rendererAddress: Long) {
+        cppDrawSkia(cppPointer, rendererAddress)
```
umberto-sonnino, line 190: So, yeah, we could have both functions here:
```kotlin
fun drawSkia(rendererAddress: Long, fit: Fit, alignment: Alignment)
fun drawSkia(rendererAddress: Long)
```
With the only caveat that the underlying JNI call must have a unique name (I think).

cpp/include/bindings/bindings_renderer.hpp
umberto-sonnino: Mmh, since we're rendering to a surface, the fact that it "knows" how big it is should be ok.
We can eventually clean it up and get it all from the `EGLThreadState`, if needed.

umberto-sonnino (approved):
LGTM!
We could keep the old draw call around just by having a different overload. In the end, our examples are pretty comprehensive for showing how to get things rolling in various use cases anyway, it's not a big deal.

app/src/main/java/app/rive/runtime/example/LowLevelActivity.kt
mjtalbot: interesting perspective, i've kinda come from the assumption that this is the control our users are going to need with the custom loop. Sure there are some usecases where what the current drawSkia does is enough, but only when you want to draw the the full drawing surface. 

We can totally keep a helper function that does parts of this, i wonder if that gives the impression that people shouldn't mess with this though piece though, but might be tunnelvisioning here a little bit

umberto-sonnino (approved):
lgtm!

PR #208, opened 2021-12-07
Rendering Metrics

### LINKED ISSUE #192: Update MetrictsActivity to show fps

umberto-sonnino:
Fixes #192 and re-introduces the `RendererMetrics` for those eager to try it.
This utility is available for API > 24, and it will Log a few Metrics every 30 frames. The exact number is hardcoded atm, but we can tweak it as needed.

There are also a few other tweaks and cleanups for:
-  some fields and methods are now private
- `JNIRendererSkia` constructor delegates init to a couple of methods
- made a few methods `const` where possible (generally good for compiler optimizations)


mjtalbot (approved):
lgtm

PR #210, opened 2021-12-07
chore: step back core-ktx and appcompat to allow for a lower compileS…

### LINKED ISSUE #209: Investigate SDK requirements

mjtalbot:
…DKVersion temporarily

fixes #209

i think we'll want to step these versions up again at some point, but right now this allows our app consumers to use sdk30 to compile.


umberto-sonnino (approved):
lgtm!
I guess tests didn't like the downgrade?

PR #213, opened 2021-12-09
Avoid multiple callbacks

### LINKED ISSUE #211: auto advance state machines on play

### LINKED ISSUE #212: fix multiple advance issue

mjtalbot:
adding a couple of test to harden how we deal with state machine play. and fix an issue where a view can receive multiple frame callbacks.

fixes #212 
fixes #211


kotlin/src/androidTest/java/app/rive/runtime/kotlin/core/RiveStateMachineStateResolutionTest.kt
```diff
@@ -0,0 +1,160 @@
...
+import app.rive.runtime.kotlin.RiveAnimationView
+import app.rive.runtime.kotlin.test.R
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+
+@RunWith(AndroidJUnit4::class)
+class RiveStateMachineStateResolutionTest {
```
umberto-sonnino, line 14: Awesome!

mjtalbot: there are a couple of things i'm not 100% on. but i guess we have tests so if we want to change things, we can... . right now i've made it so just the default loading up of a rive file with autoplay on does not settle the state machine state. ... it all feels so edge casy... (largely because that interferes with how reset behaves, ... ... but .. jeee it gets really picky :D  could track something there too though to let people override things. ... honestly i think im in too deep and :D )

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -348,9 +348,11 @@ open class RiveAnimationView(context: Context, attrs: AttributeSet? = null) :
     fun play(
         animationName: String,
         loop: Loop = Loop.AUTO,
-        direction: Direction = Direction.AUTO, isStateMachine: Boolean = false
+        direction: Direction = Direction.AUTO,
+        isStateMachine: Boolean = false,
+        settleInitialState: Boolean = true
     ) {
-        renderer.play(animationName, loop, direction, isStateMachine)
+        renderer.play(animationName, loop, direction, isStateMachine, settleInitialState)
```
umberto-sonnino, line 355: I guess this is our resolver!

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -87,6 +85,19 @@ open class RiveArtboardRenderer(
         }
     }
 
+    fun _advanceStateMachineInstance(
+        stateMachineInstance: StateMachineInstance,
```
umberto-sonnino, line 89: Kotlin doesn't really like the _ at the start, can't we just make it private?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
mjtalbot: yea, forces a 0 advance on the state machine instance

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -154,19 +166,20 @@ open class RiveArtboardRenderer(
         loop: Loop = Loop.AUTO,
         direction: Direction = Direction.AUTO,
         isStateMachine: Boolean = false,
+        settleInitialState: Boolean = true,
```
umberto-sonnino, line 169: Oh, we're defaulting to true? Is that so our users get "what they expect"?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
mjtalbot: dont see any reason why not :D i wonder if there are some warnings i was ignoring here..

kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererSkia.kt
```diff
@@ -36,6 +36,7 @@ abstract class RendererSkia(trace: Boolean = false) :
 
     var isPlaying: Boolean = false
         private set
+    var lastFrameTimeNanos: Long = 0L
```
umberto-sonnino, line 39: This private too?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
mjtalbot: this is so that by default, when you say 

play("statmachine")
trigger("foo")

and foo will only do something if we're at the settled state... it works. yeah.. the thought is that for the high level thing this is intended behaviour

kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererSkia.kt
mjtalbot: lets make sure they cant get our nanos!

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -330,10 +344,24 @@ open class RiveArtboardRenderer(
...
             stateMachines.add(stateMachineInstance)
         }
+
+        // Special case:
+        // When we start to "play" a state machine, we want it to settle on its initial state
+        // otherwise it maybe "stuck" on the Enter state causing issues for fireState triggers.
+        // https://2dimensions.slack.com/archives/CLLCU09T6/p1638984141105200
+        if (settleStateMachineState) {
+            activeArtboard?.let {
+                _advanceStateMachineInstance(stateMachineInstance, it, 0f)
```
umberto-sonnino: Oh, cool, so we get notified about this too, makes sense!

mjtalbot: yeah exactly.. our tests caught this :D

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
umberto-sonnino: Yeah I think I saw a few warnings about this in Android Studio...just nitpicks really

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
umberto-sonnino: Fair fair, yeah, bit of a shame that we're sprinkling our own code with `settleInitialState = false`, but this is going to make it so much easier for our users.

kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererSkia.kt
umberto-sonnino: Nanos want their privacy

umberto-sonnino (approved):
LGTM!
Should we have something in our README about this new flag?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
mjtalbot: yeah i'm not really in love with it, but it does get us closer to what the editor is doing. 

I didnt a good way to do this without passing it through, i guess I could have made a class var or a "rive" setting that determines this

PR #214, opened 2021-12-09
update autoplay behaviour for state machine initialisation

mjtalbot:
i think this is more consistent with our approach of resolving the state machine when we play by default. 

you can still control it, just disable autoplay (i think the way i had it defaulted play to resolving, and autoplay to not resolving which is weird)

### DIFF
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
@@ -449,11 +449,11 @@ open class RiveArtboardRenderer(
                 play(animationName = it)
             } ?: run {
                 stateMachineName?.let {
-                    // only settle initial state on explicit play call
-                    play(animationName = it, isStateMachine = true, settleInitialState = false)
+                    // With autoplay, we default to settling the initial state
+                    play(animationName = it, isStateMachine = true, settleInitialState = true)
                 } ?: run {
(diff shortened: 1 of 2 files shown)


kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -449,11 +449,11 @@ open class RiveArtboardRenderer(
...
                 stateMachineName?.let {
-                    // only settle initial state on explicit play call
-                    play(animationName = it, isStateMachine = true, settleInitialState = false)
+                    // With autoplay, we default to settling the initial state
+                    play(animationName = it, isStateMachine = true, settleInitialState = true)
                 } ?: run {
-                    // only settle initial state on explicit play call
-                    play(settleInitialState = false)
+                    // With autoplay, we default to settling the initial state
+                    play(settleInitialState = true)
```
umberto-sonnino, line 456: 👍 
I'm guessing specifying the param is there for clarity. We're already defaulting to true, aren't we?

umberto-sonnino (approved):

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
mjtalbot: we are already defaulting to true! I totally did not do it for clarity, but i guess its pretty clear now, and I think thats a good thing

PR #216, opened 2021-12-16
CMake Integration

umberto-sonnino:
Add CMakeLists.txt to support building `.so`s with CMake, integrate C++ with Android Studio, builds c++ static library deps from CMake directly.
Depends on https://github.com/rive-app/rive-cpp/pull/131


cpp/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
...
+  src/bindings/bindings_layer_state.cpp
+  src/bindings/bindings_aabb.cpp
+  src/bindings/bindings_helper.cpp
+  src/helpers/general.cpp
+  src/helpers/thread.cpp
+  src/helpers/egl_thread_state.cpp
+  src/helpers/settings.cpp
+  )
+
+# -I$(LIBRIVE)/include \
```
mjtalbot, line 44: should this just get deleted?, or does it serve a purpose going forward?

kotlin/build.gradle
```diff
@@ -6,18 +6,30 @@ apply plugin: 'org.jetbrains.dokka'
 android {
     compileSdkVersion 31
     buildToolsVersion "30.0.3"
+    ndkVersion "22.1.7171670"
```
mjtalbot, line 9: 👀  cool, we just need to make sure we stay synced on the ndk i guess (but i think we got there now)

kotlin/build.gradle
```diff
@@ -6,18 +6,30 @@ apply plugin: 'org.jetbrains.dokka'
...
+            }
+        }
     }
 
     kotlinOptions {
         jvmTarget = JavaVersion.VERSION_1_8.toString()
     }
+    externalNativeBuild {
+        cmake {
+            path file('../cpp/CMakeLists.txt')
```
mjtalbot, line 29: ohh exciting

mjtalbot (approved):
lgtm, dont really know what its doing. but it looks promising. happy to give this a spin on my machine as well to make sure it works across devices

PR #219, opened 2021-12-17
Fix FPS counter

### LINKED ISSUE #218: Fix static fps counter in `MetricsActivity`

umberto-sonnino:
Also, some housekeeping for `JNIRendererSkia`.

Fixes #218

### DIFF
--- a/cpp/include/helpers/tracer.hpp
+++ b/cpp/include/helpers/tracer.hpp
@@ -1,8 +1,10 @@
 #ifndef _RIVE_ANDROID_TRACER_HPP_
 #define _RIVE_ANDROID_TRACER_HPP_
 
-#include "helpers/general.hpp"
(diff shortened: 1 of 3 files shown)


mjtalbot (approved):
lgtm. 

with the refactor its quite hard to see what changed

does this fix working out fps for multiple animations at the same time?

umberto-sonnino:
> lgtm.
> 
> with the refactor its quite hard to see what changed
> 
> does this fix working out fps for multiple animations at the same time?

It does! 
Sorry about that, the second commit is the one with the fix, that one should be easier to read! It's just about moving the FPS variables from static into a class field so that every renderer has its own copy. The next level would be to have something similar to our Tracer so that we can attach objects that do their own calculations on it depending on top-level configuration

PR #223, opened 2022-01-11
Maint: Update to latest rive-cpp

zplata:
Updating to latest `rive-cpp` commit: https://github.com/rive-app/rive-cpp/commit/4b7ce871ca5e2d11a0bd9ef7e281722d4d12f7db


mjtalbot (changes requested):
Looks great! 

there's been another commit on rive-cpp though (https://github.com/rive-app/rive-cpp/commits/master) so i think we might as well update to that commit, build the .so's again and update this pr to that. and then merge!!

mjtalbot (approved):
lgtm!

PR #225, opened 2022-01-26
Updating rive-android for changes to rive-cpp.

luigi-rosso:
We changed rive-cpp so that it can build regular skia and the rive optimized (slimmed down) skia. We do this by building two versions of skia, so anything depending on the optimized one needs to update references. Lots of this was done in rive-app/rive-cpp but some needs to done in the various runtimes. I think this is all that's necessary here but please take a look!


mjtalbot (changes requested):
looks great, we need to build the .so's otherwise this doesn't change anything :P 

will throw a ticket in for it.

mjtalbot (approved):
lgtm

PR #228, opened 2022-02-01
fix: update start to only trigger frame schedules once per current fr…

mjtalbot:
…ametime

### DIFF
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererSkia.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererSkia.kt
@@ -1,5 +1,6 @@
 package app.rive.runtime.kotlin.renderers
 
+import android.util.Log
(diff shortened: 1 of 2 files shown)


kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RendererSkia.kt
```diff
@@ -56,15 +68,38 @@ abstract class RendererSkia(trace: Boolean = false) :
...
         cppStop(cppPointer)
     }
 
+
+    /**
+     * Calls stop, and removes any pending frameCallbacks from the Choreographer
+     *
+     * NOTE: this is NOT safe to call from the animation thread.
+     * e.g inside .draw() / .advance(elapsed: Float) callbacks
+     */
```
umberto-sonnino, line 96: 🙌

umberto-sonnino (approved):

PR #230, opened 2022-02-02
Initialize pointers to null

umberto-sonnino:
Fixes a crash with invisible Views. JNIRenderer was initialized but its corresponding window surface wasn't. The destructor would almost always try to delete a missing `mSkRenderer`.


mjtalbot (approved):
lgtm

PR #232, opened 2022-02-04
Update cpp & rebuild libs

umberto-sonnino:
Fixes non-rendering open paths & bumps cpp submodule to the latest


mjtalbot (approved):
lgtm. i think the .so change is kinda stealth but whatever i guess.. i think its down to how we set the optimized skia branch

umberto-sonnino:
> dumb question. did we update rive-cpp to point to the correct optimized skia? or did we just bump contents of the rive-skia branch that we point to?

I swapped our fork's branches so that `rive` is now pointing to yesterday's update, and I renamed our old `rive` to be `rive_Dec_20`, so the change is transparent in terms of our scripts.

PR #233, opened 2022-02-10
fix: make sure we update the display if we change the fit/alignment r…

mjtalbot:
…ules

fixes an issue we ran into getting react-native up to speed. 

basically when you change fit/alignment we don't neccessarily draw another frame to update the display. this forces that to happen.


kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -6,8 +6,8 @@ import app.rive.runtime.kotlin.renderers.RendererSkia
 
 open class RiveArtboardRenderer(
     // PUBLIC
-    var fit: Fit = Fit.CONTAIN,
-    var alignment: Alignment = Alignment.CENTER,
+    var _fit: Fit = Fit.CONTAIN,
```
mjtalbot, line 9: better way to do this?

zplata (approved):

PR #234, opened 2022-02-11
Recycle my day

mjtalbot:
do not merge. (obvs) 

its what we've been doing to figure out whats happening with our egl contexts. 

problem: we run out of surfaces after 200are drawn, and then we need to restart the app to be able to render rive animations

check the recycler view, and scroll down to the bottom and then up a bit to see this in action

this is not recycler view specific! (you can repro it by just oopening and closing the other animation pages)

we've checked threads, they get joined and disappear fine. 

we think its somewhere in egl_thread_state.cpp and that we are not  clearing up our surfaces properly. but we've not hit gold yet, no smoking gun.


umberto-sonnino:
On Monday, I'll try to see if this is reproducible in some other way without our setup, maybe even just using EGL in Java.
I'll try also to use the `-d` debug compile flag to optimize link times.

umberto-sonnino:
> Can you link the Google project sample for rendering to surfaces? @umberto-sonnino ... (I'll do it if I can find the link)

[Here](https://android.googlesource.com/platform/frameworks/opt/gamesdk/+/refs/heads/sdk-release/) you go!

PR #236, opened 2022-02-16
bugfix: add synchronized to avoid list operations that are not thread…

mjtalbot:
… safe in our high level renderer

this is the "lock" approach  

note: re-entrant lock which should not be meaningful here, just more of a just in case. 
note: no tryLock -> skip approach (I've had that in before, but that would mess up timings over time)

I've not been able to reproduce the race conditions in a normal setting.

But creating an activity that calls view.stop & play every 10 nanoseconds in a subthread does the trick as long as you comment out `stopThread()` in RiveArtboardRenderer.kt (which runs you into looper issues). 

```kotlin
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.concurrent)
        var view = findViewById<RiveAnimationView>(R.id.simple_view_asset)

        var count = 0

        thread {
            while (true) {
                Thread.sleep(0, 10)
                if ((count) % 2 == 0) {
                    view.stop()
//                    view.pause()
                } else {
                    view.play()
                }
                count++
            }
            true
        }

    }
```

performance wise this seems to be in the ballpark of what we have previously had, the reading I get from our Metrics activity change day by day. But right now I'm getting no locks to settle on 4.40ms and with locks to settle on 4.39ms after a few minutes. but yesterday I settled on 4.44ms, & 4.50ms, so there is certainly some other variation to take into account here.

### DIFF
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
@@ -4,6 +4,7 @@ import app.rive.runtime.kotlin.core.*
 import app.rive.runtime.kotlin.core.errors.ArtboardException
 import app.rive.runtime.kotlin.renderers.RendererSkia
 import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
 
 open class RiveArtboardRenderer(
     // PUBLIC
@@ -28,7 +29,7 @@ open class RiveArtboardRenderer(
     private var playingAnimationList: HashSet<LinearAnimationInstance> = HashSet()
     val playingAnimations: HashSet<LinearAnimationInstance>
         public get() {
-            return synchronized(lock) {
+            return lock.withLock {
                 playingAnimationList.toHashSet()
             }
         }
@@ -36,7 +37,7 @@ open class RiveArtboardRenderer(
     private var playingStateMachineList: HashSet<StateMachineInstance> = HashSet()
     val playingStateMachines: HashSet<StateMachineInstance>
         public get() {
-            return synchronized(lock) {
+            return lock.withLock {
                 playingStateMachineList.toHashSet()
             }
         }
@@ -44,7 +45,7 @@ open class RiveArtboardRenderer(
     private var animationList = mutableListOf<LinearAnimationInstance>()
     val animations: List<LinearAnimationInstance>
         public get() {
-            return synchronized(lock) {
+            return lock.withLock {
                 animationList.toList()
             }
         }
@@ -53,7 +54,7 @@ open class RiveArtboardRenderer(
     private var stateMachineList = mutableListOf<StateMachineInstance>()
     val stateMachines: List<StateMachineInstance>
         public get() {
-            return synchronized(lock) {
+            return lock.withLock {
                 stateMachineList.toList()
             }
         }
@@ -89,7 +90,7 @@ open class RiveArtboardRenderer(
     }
 
     override fun advance(elapsed: Float) {
-        synchronized(lock) {
+        lock.withLock {
             activeArtboard?.let { ab ->
                 // animations could change, lets cut a list.
                 // order of animations is important.....
@@ -123,7 +124,7 @@ open class RiveArtboardRenderer(
             }
             // Are we done playing?
             if (!hasPlayingAnimations) {
-                stopThread()
+//                stopThread()
             }
         }
     }
@@ -176,7 +177,7 @@ open class RiveArtboardRenderer(
     }
 
     fun clear() {
-        synchronized(lock) {
+        lock.withLock {
             playingAnimationList.clear()
             animationList.clear()
             playingStateMachineList.clear()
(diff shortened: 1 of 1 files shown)


kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -148,20 +176,24 @@ open class RiveArtboardRenderer(
...
     fun reset() {
-        stopAnimations()
-        stop()
-        clear()
-        selectedArtboard?.let {
-            setArtboard(it)
+        synchronized(lock) {
+            stopAnimations()
+            stop()
+            clear()
```
umberto-sonnino, line 191: If `clear` is already synchronized, do we need to lock here also?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -24,18 +25,44 @@ open class RiveArtboardRenderer(
...
-    var playingAnimations = HashSet<LinearAnimationInstance>()
-        private set
-    var playingStateMachines = HashSet<StateMachineInstance>()
-        private set
+    private var playingAnimationList: HashSet<LinearAnimationInstance> = HashSet()
+    val playingAnimations: HashSet<LinearAnimationInstance>
+        public get() {
+            return synchronized(lock) {
+                playingAnimationList.toHashSet()
+            }
```
umberto-sonnino, line 33: Do we always need reads to be synchronized?

umberto-sonnino: Thinking about this, I guess we do! The original report was on `toList()`

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -183,121 +217,147 @@ open class RiveArtboardRenderer(
...
+            val stateMachineInstances = _getOrCreateStateMachines(stateMachineName)
+            stateMachineInstances.forEach {
+                (it.input(inputName) as SMINumber).value = value
+                _play(it, settleStateMachineState = false)
+            }
         }
     }
 
-    // PRIVATE FUNCTIONS
+// PRIVATE FUNCTIONS
```
umberto-sonnino, line 360: Mmh, so we're `synchronize`-ing all our public APIs and leaving the private ones unflitered?
Our private functions below contain a few read/write ops that could be considered critical sections.
Shouldn't we `synchronize` those instead? Keep our critical sections minimal

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
mjtalbot: its reentrant, tbf i just locked all public methods, probably not

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
mjtalbot: yeah i had just playingAnimations wrapped around a lock before, but that wasn't actually enough as i guess operations on it werent ...

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
mjtalbot: honestly, I felt like it was simpler to wrap our public methods and advance. 

our public methods are fast as they just end up resolving to creating an animation instance and adding it to a list and maybe a hash set.

mjtalbot: also going forward it felt less error prone to go with "you changing something? cool grab a lock, you can change stuff" rather than having to contend with thread safety every time we touch this (especially as we dont have tests that catch this rn)

mjtalbot: this is also closer to how i envision the queuing mechanism to work, queue some actions, the thread will get to your actions before each frame

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -308,13 +346,13 @@ open class RiveArtboardRenderer(
     }
 
     private fun _animations(animationNames: Collection<String>): List<LinearAnimationInstance> {
-        return animations.filter { animationInstance ->
+        return animationList.filter { animationInstance ->
```
mjtalbot, line 349: i wonder if we should have these use animations, probably (considering toList is not thread safe, guessing this isnt either @umberto-sonnino?)

umberto-sonnino (approved):
This looks pretty darn clean with sycn'd collections!

PR #239, opened 2022-02-22
Thread pool

umberto-sonnino:
Re-evaluates the way we handle threading.
Namely, we introduce a thread pool to retain EGL Contexts to circumvent an Adreno bug that prevents us from instantiating more than 200 of them (also reported here: [Google Issue Tracker](https://issuetracker.google.com/issues/202616723))

This new setup introduces a number of potential race conditions:
1. Since the thread is retaining the context, JNIRendererSkia cannot be its owner anymore. The owner is, in fact, a Thread Pool Singleton that distributes threads as needed. When a thread isn't available, a new one is created.
  - This means that `JNIRendererSkia` *cannot* delete the thread, but it can only release it to the pool once finished
2. The Thread is then responsible for deleting JNIRendererSkia (which otherwise would just be leaking memory). Since the thread is executing work from a work queue and the work queue is potentially filled when the thread is released, we cannot delete the JNIRendererSkia pointer right away. Thus the thread will:
  - signal that it is done working, preventing any more work to be added to the queue
  - empty the work queue and clean up the references to JNI/Kotlin objects that are tied to the View that is being destroyed
  - force one more task onto the work queue: that is a callback to delete JNIRendererSkia once it's done running through all the remaining tasks

Following this logic, we prevent referencing null pointers. 
That is also valid on the Kotlin side, where we set our `cppPointer` to 0 immediately after scheduling the JNIRendererSkia for deletion. To prevent using this reference again, a few 0-checks have been added to gate access to callbacks (`advance()` and `draw()`, which are called by our rendering threads from C++), and `start()`/`stop()` helper functions.

I think there's still potentially quite a lot of cleanup to do, but this solves our problems with `EGLContexts` for now.


app/src/main/AndroidManifest.xml
```diff
@@ -23,7 +23,10 @@
                 android:value="androidx.startup" />
         </provider>
 
-        <activity android:name=".MainActivity">
+        <activity
+            android:name=".MainActivity"
+            android:clearTaskOnLaunch="true"
```
mjtalbot, line 28: is this something we should always have had?

app/src/main/java/app/rive/runtime/example/RecyclerActivity.kt
```diff
@@ -0,0 +1,77 @@
...
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_recycler)
+
+
+        val recyclerView: RecyclerView = findViewById(R.id.recycler_view)
+        recyclerView.adapter = RiveAdapter()
+    }
+}
+
+class FileCheater(val number: Int, val resource: Int)
```
mjtalbot, line 26: 👀  deserves a comment :P

mjtalbot: oh hold on a second. this is my terrible work :D

mjtalbot: still deserves a comment lol

app/src/main/java/app/rive/runtime/example/RecyclerActivity.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    override fun areContentsTheSame(oldItem: FileCheater, newItem: FileCheater): Boolean {
+        return oldItem.number == oldItem.number
+    }
+}
+
+class RiveAdapter : ListAdapter<FileCheater, RiveAdapter.RiveViewHolder>(RiveFileDiffCallback) {
+
+    /* ViewHolder for Flower, takes in the inflated view and the onClick behavior. */
```
mjtalbot, line 40: its not flowers, its file cheaters

app/src/main/res/layout-w1240dp/activity_recycler_item.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.core.widget.NestedScrollView xmlns:android="http://schemas.android.com/apk/res/android"
```
mjtalbot, line 2: whats up with the additional layouts for different dps' should we just axe em? (assuming this is all crap i committed btw)

app/src/main/res/layout/main.xml
```diff
@@ -94,6 +94,12 @@
...
                 android:text="Assets" />
+
+            <Button
+                android:id="@+id/go_recycler"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:text="Recycler" />
         </LinearLayout>
 
     </ScrollView>
```
mjtalbot, line 105: can probably get rid of paff.riv :P

mjtalbot: or name it something better, like artistic oval quickly shifting to the top right

cpp/include/helpers/egl_thread_state.hpp
```diff
@@ -98,13 +101,22 @@ namespace rive_android
...
 		static void* getProcAddress(const char*);
-		void clearSurface();
 		bool configHasAttribute(EGLConfig, EGLint, EGLint) const;
 
 		EGLBoolean makeCurrent(EGLSurface surface) const
 		{
 			return eglMakeCurrent(mDisplay, surface, surface, mContext);
 		}
+
+		sk_sp<GrDirectContext> getGrContext()
```
mjtalbot, line 111: probably makes sense, but should mSkContext be called something like mGrContext? or should getGrContext be called getSkContext?

mjtalbot: or are we implementing a shared interface?

cpp/include/helpers/worker_thread.hpp
```diff
@@ -57,13 +60,49 @@ namespace rive_android
 		{
 			std::lock_guard<std::mutex> threadLock(mThreadMutex);
 			terminateThread();
+			// Detach thread from the JVM.
+			detachThread();
 		}
 
-		void run(Work work)
+		bool run(Work work)
```
mjtalbot, line 67: work work

cpp/include/helpers/worker_thread.hpp
```diff
@@ -89,6 +128,14 @@ namespace rive_android
 			mThread.join();
```
mjtalbot, line 128: when do these threads stop now?

cpp/include/helpers/worker_thread.hpp
```diff
@@ -130,11 +177,45 @@ namespace rive_android
...
 		std::queue<std::function<void(ThreadState*)>>
 		    mWorkQueue GUARDED_BY(mWorkMutex);
 		std::condition_variable_any mWorkCondition;
 	};
 
+	class ThreadManager
+	{
+	private:
+		ThreadManager() : mThreadPool{} {};
+		~ThreadManager()
```
mjtalbot, line 193: ah so when the manager goes, the thread show is over

cpp/src/helpers/egl_thread_state.cpp
```diff
@@ -155,7 +171,8 @@ namespace rive_android
 
 		GrBackendRenderTarget backendRenderTarget(
 		    mWidth, mHeight, 1, 8, fbInfo);
-		static SkSurfaceProps surfaceProps(0, kUnknown_SkPixelGeometry);
+		static SkSurfaceProps surfaceProps(SkSurfaceProps::kDynamicMSAA_Flag,
```
mjtalbot, line 174: MSAA better than 0?

cpp/src/helpers/worker_thread.cpp
```diff
@@ -0,0 +1,48 @@
...
+		if (mInstance == nullptr)
+		{
+			mInstance = new ThreadManager();
+		}
+		return mInstance;
+	}
+
+	WorkerThread<EGLThreadState>* ThreadManager::acquireThread(const char* name)
+	{
+		std::lock_guard<std::mutex> threadLock(mMutex);
```
mjtalbot, line 21: whats this do?

mjtalbot: (and if it gets a lock, why does it not release a lock?)

mjtalbot (approved):
"lgtm" but i think its a rubber stamp.

Before merging this, can we run the test app using this on the device farm? 
Do you think any manual testing is needed before merging this? 

Should we do some memory checks over time to make sure we don't leak memory? (i guess just a matter of creating and destroying a boat load of views with rive animations and make sure once we back out its all gone?)

also do we actually set a limit anywhere for number of Threads? or we just aggressively reuse threads, so hopefully we dont hit a limit (which seems sensible)

app/src/main/AndroidManifest.xml
umberto-sonnino: Android Studio flagged it as something that needed to be there, so I just accepted it!

app/src/main/res/layout/main.xml
umberto-sonnino: paff is now part of my debugging routine, it'll be sad to see it go

cpp/include/helpers/egl_thread_state.hpp
umberto-sonnino: We make a GL context from a `GrDirectContext`, but the field is called `mSkContext` since it's the Skia context. 
Yeah the naming here could be ironed out a bit.  😄

umberto-sonnino:
> "lgtm" but i think its a rubber stamp.
> 
> Before merging this, can we run the test app using this on the device farm? Do you think any manual testing is needed before merging this?

You're totally right, I'll give it a spin out there and see!

> Should we do some memory checks over time to make sure we don't leak memory? (i guess just a matter of creating and destroying a boat load of views with rive animations and make sure once we back out its all gone?)
> 

I did do some manual testing around threads getting reused. There is a framework called googletest for running native tests within the NDK. I haven't spent the time to set it up but we should most definitely do it!

> also do we actually set a limit anywhere for number of Threads? or we just aggressively reuse threads, so hopefully we dont hit a limit (which seems sensible)

On this, yeah, I haven't set a hard limit on the number of threads but left it as a potential `// TODO:`, since stuff is getting aggressively reused, we shouldn't see massive amounts of thread getting instantiated. We still have a hard limit on the number of EGLContexts, which means that we have a hard `200` limit on the number of working threads we can have. 

Another improvement for this setup is to have our caching work with a TTL, so that we don't keep threads around infinitely, but we let them go once a certain amount of time has passed.

app/src/main/res/layout/main.xml
mjtalbot: 🤣

cpp/src/helpers/worker_thread.cpp
mjtalbot: omg. pro locks! thats super cool

PR #240, opened 2022-02-28
Improve logging for debug and add EGL error checks

umberto-sonnino:
Just some QoL improvements that I had left in the other experimental branch.
These changes don't affect our `.so`s since they are debug-only.


mjtalbot (approved):
lgtm! nice to get these pieces in!

PR #251, opened 2022-05-04
Add small fix for release workflow to avoid pr closing edge case

zplata:
So that these publish flows only get run when PR's to master are closed and merged


umberto-sonnino (approved):
LGTM! 
We're doing the same on iOS I see.

PR #253, opened 2022-05-10
Add touch events

mjtalbot:
does what it says on the tin. adds eamples like we have in rive-ios

https://user-images.githubusercontent.com/1216025/167687558-aaf4e813-7bba-402a-9444-8cb9e0af9a97.mov

I've tried getting some basic tests in as well. still working on a test that shows we convert between different world spaces properly


kotlin/src/androidTest/java/app/rive/runtime/kotlin/core/RiveStateMachineTouchEventTest.kt
```diff
@@ -0,0 +1,132 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import android.os.SystemClock
+import android.view.MotionEvent
+
+
+@RunWith(AndroidJUnit4::class)
+class RiveStateMachineTouchEventTest {
```
mjtalbot, line 16: @zplata @duncandoit example of basic tests for touch events

cpp/src/bindings/bindings_helper.cpp
```diff
@@ -30,6 +30,32 @@ extern "C"
...
+		auto touchSpaceBounds = (rive::AABB*)touchSpaceBoundsRef;
+		auto touchSpaceLocation = (rive::Vec2D*)touchSpaceLocationRef;
+		auto artboardSpaceLocation = (rive::Vec2D*)artboardSpaceLocationRef;
+
+		rive::Mat2D forward = rive::computeAlignment(
+		    fit, alignment, *touchSpaceBounds, *artboardSpaceBounds);
+		rive::Mat2D inverse = forward.invertOrIdentity();
+
+		rive::Vec2D convertedLocation = inverse * *touchSpaceLocation;
+		*artboardSpaceLocation += convertedLocation;
```
mikerreed, line 56: curious: why are we adding instead of just assigning?

cpp/src/bindings/bindings_vec2d.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "rive/math/vec2d.hpp"
+#include <jni.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+	// AABB
+
+	JNIEXPORT jlong JNICALL Java_app_rive_runtime_kotlin_core_Vec2D_constructor(
```
mikerreed, line 12: curious: why wrap this at all? Seems like a lot to just pass around 2 floats. Can we use Android's point (or just an array)?

kotlin/src/main/java/app/rive/runtime/kotlin/core/Vec2D.kt
```diff
@@ -0,0 +1,40 @@
...
+ * Representation of a 2D Vector
+ *
+ * This object has a counterpart in c++, which implements a lot of functionality.
+ * The [cppPointer] keeps track of this relationship.
+ *
+ * The Vec2D helps us describe and keep track of x,y coordinates
+ *
+ */
+class Vec2D {
+    var cppPointer: Long
```
mikerreed, line 13: just 2 floats?

mikerreed (approved):
... also OK if we want to land as is -- and discuss small classes like Vec2D separately

cpp/src/bindings/bindings_helper.cpp
mjtalbot: Becauseeeeeee I'm doing a hack where I create the object in kotlin, but then set it in CPP. Yeah this sucks though. Could just have this return a pointer to a new vec2d ... Or bypass vec2d entirely just gotta dance around the jni

cpp/src/bindings/bindings_vec2d.cpp
mjtalbot: Probably can! Will give it a try, and maybe but @umberto who's a bit better versed here!

cpp/src/bindings/bindings_helper.cpp
umberto-sonnino: Yeah, this function could return a `Float[2]`, not sure if JNI makes it a nightmare though.
Do we care about any other value other than `artboardSpaaceLocation`?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -213,6 +213,15 @@ open class RiveAnimationView(context: Context, attrs: AttributeSet? = null) :
         renderer.targetBounds = AABB(width.toFloat(), height.toFloat())
     }
 
+    override fun onSurfaceTextureAvailable(
+        surfaceTexture: SurfaceTexture,
+        width: Int,
+        height: Int
+    ) {
+        super.onSurfaceTextureAvailable(surfaceTexture, width, height)
+        renderer.targetBounds = AABB(width.toFloat(), height.toFloat())
```
umberto-sonnino, line 222: Uh interesting, was this creating problems?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveTextureView.kt
```diff
@@ -6,10 +6,12 @@ import android.content.ContextWrapper
 import android.graphics.SurfaceTexture
 import android.util.AttributeSet
 import android.util.Log
+import android.view.MotionEvent
```
umberto-sonnino, line 9: Extra imports?

cpp/src/bindings/bindings_helper.cpp
mjtalbot: coolios, just trying to do the dance with a PointF atm. but if that doesnt work out, float[2] sounds great too

kotlin/src/main/java/app/rive/runtime/kotlin/RiveTextureView.kt
mjtalbot: oh must be debugging leftovers

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
mjtalbot: yeah, we didn't have targetBounds set initially! so we were typically addressing a 0,0 AABB for that!

cpp/src/bindings/bindings_helper.cpp
mjtalbot: replaced with PointF, we're passing in the ref to it, and just settings its value so we dont have to think too hard about garbage collection

PR #257, opened 2022-05-26
Roll cpp, update gradle

mikerreed:
Do we care about Gradle version? When I was building, Android Studio suggested I upgrade ...


mjtalbot (approved):
lgtm. 

i think we should try to keep those things up to date with whatever the build tools unless we have a specific reason not to. if we run into problems, its probably something we should learn about

PR #265, opened 2022-06-23
Build remotely

mjtalbot:
testing building... 

still early days. trying to make the caching stuff a bit more reusable. i think i can use the stuff here in the recorder scripts i went for yesterday


cpp/.ndk_version.bots
```diff
@@ -0,0 +1 @@
+r22b
```
umberto-sonnino, line 1: Is this the filename? I thought the version number would always be semver'd

cpp/cache_helper.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo "Grabbing cached build from $ARCHIVE_URL"
+    mkdir -p $SKIA_RECORDER_DIR
+    curl --output $SKIA_RECORDER_DIR/$ARCHIVE_FILE_NAME $ARCHIVE_URL 
+    pushd $SKIA_RECORDER_DIR
+    tar -xf $ARCHIVE_FILE_NAME out include archive_contents third_party
+}
+
+is_build_cached_remotely() {
+    echo "Checking for cache build $ARCHIVE_URL"
+    if curl --output /dev/null --head --silent --fail $ARCHIVE_URL
```
umberto-sonnino, line 68: Oh, so this gets cached between one build and the other. How long does this cache live for?

mjtalbot: forever

mjtalbot: (or until we chose to delete it)

PR #286, opened 2023-03-15
Virtualization fixes

umberto-sonnino:
- Adds a `ViewPager` example with Horizontal scrolling
- Fixes a few issues with timers being unreliable
- Does not reload the file if already loaded
- Fix `disposeDependencies()` visibility and properly clean up the `file` when calling it in the subclass
- Add a few fields to `DetachedState` so it's more resilient and it can restore the right animation
- Adds a tweaks to the `RecyclerView` example

I'm curious to see if this addresses also those reports about `RecyclerView` from our issue tracker, from my testing this seems more resilient now.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -2,13 +2,13 @@ apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdkVersion 31
+    compileSdkVersion 33
     buildToolsVersion "30.0.3"
 
     defaultConfig {
         applicationId "app.rive.runtime.example"
         minSdkVersion 21
-        targetSdkVersion 30
+        targetSdkVersion 33
         versionCode 1
         versionName "1.0"
 
@@ -36,6 +36,9 @@ android {
 
     kotlinOptions {
         jvmTarget = JavaVersion.VERSION_1_8.toString()
+    }
+    buildFeatures {
+        viewBinding true
     }
 }
 
--- a/app/src/main/java/app/rive/runtime/example/RecyclerActivity.kt
+++ b/app/src/main/java/app/rive/runtime/example/RecyclerActivity.kt
@@ -1,5 +1,6 @@
 package app.rive.runtime.example
 
+import android.graphics.Color
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
@@ -27,11 +28,11 @@ data class RiveFileResource(val number: Int, val resource: Int)
 
 object RiveFileDiffCallback : DiffUtil.ItemCallback<RiveFileResource>() {
     override fun areItemsTheSame(oldItem: RiveFileResource, newItem: RiveFileResource): Boolean {
-        return oldItem.number == oldItem.number
+        return oldItem.number == newItem.number
     }
 
     override fun areContentsTheSame(oldItem: RiveFileResource, newItem: RiveFileResource): Boolean {
-        return oldItem.number == oldItem.number
+        return oldItem.number == newItem.number
     }
 }
 
@@ -59,6 +60,12 @@ class RiveAdapter :
 
     override fun onBindViewHolder(holder: RiveViewHolder, position: Int) {
         val riveFileResource = getItem(position)
+        // Alternate background colors to differentiate various elements.
+        if (position % 2 == 1) {
+            holder.itemView.setBackgroundColor(Color.parseColor("#FFFFFF"));
+        } else {
+            holder.itemView.setBackgroundColor(Color.parseColor("#FFFAF8FD"));
+        }
         holder.bind(riveFileResource)
     }
 
@@ -67,6 +74,11 @@ class RiveAdapter :
     }
 
     override fun getItem(position: Int): RiveFileResource {
-        return RiveFileResource(position, R.raw.circle_move)
+        val res = if (position % 2 == 1) {
+            R.raw.basketball
+        } else {
+            R.raw.circle_move
+        }
+        return RiveFileResource(position, res)
     }
 }
--- a/app/src/main/java/app/rive/runtime/example/ViewPagerActivity.kt
+++ b/app/src/main/java/app/rive/runtime/example/ViewPagerActivity.kt
@@ -1,0 +1,85 @@
+package app.rive.runtime.example
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.Menu
+import android.view.MenuItem
+import android.view.ViewGroup
+import androidx.appcompat.app.AppCompatActivity
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.ListAdapter
+import androidx.recyclerview.widget.RecyclerView
+import app.rive.runtime.example.databinding.ViewPagerRiveWrapperBinding
+import app.rive.runtime.example.databinding.ViewPagerBinding
+
+
+class ViewPagerActivity : AppCompatActivity() {
+    private lateinit var binding: ViewPagerBinding
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+
+        binding = ViewPagerBinding.inflate(layoutInflater)
+        setContentView(binding.root)
+
+        binding.viewPager.apply {
+            adapter = RiveViewPagerAdapter().apply {
+                submitList(
+                    listOf(
+                        R.raw.basketball,
+                        R.raw.off_road_car_blog,
+                        R.raw.flux_capacitor,
+                        R.raw.artboard_animations,
+                    )
+                )
+            }
+        }
+    }
+
+    private class RiveTestViewHolder(val binding: ViewPagerRiveWrapperBinding) :
+        RecyclerView.ViewHolder(binding.root) {
+        fun bind(res: Int) {
+            binding.riveTestView.setRiveResource(res)
+        }
+    }
+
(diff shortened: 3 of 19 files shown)


app/build.gradle
```diff
@@ -37,6 +37,9 @@ android {
     kotlinOptions {
         jvmTarget = JavaVersion.VERSION_1_8.toString()
     }
+    buildFeatures {
+        viewBinding true
```
mjtalbot, line 41: what does this do? ohhhhh.... https://developer.android.com/topic/libraries/view-binding... is this just a positive for people consuming the runtime, or does enabling this have side effects?

mjtalbot: (do we have an example that utilizes this?)

app/src/main/java/app/rive/runtime/example/RecyclerActivity.kt
```diff
@@ -27,11 +28,11 @@ data class RiveFileResource(val number: Int, val resource: Int)
 
 object RiveFileDiffCallback : DiffUtil.ItemCallback<RiveFileResource>() {
     override fun areItemsTheSame(oldItem: RiveFileResource, newItem: RiveFileResource): Boolean {
-        return oldItem.number == oldItem.number
+        return oldItem.number == newItem.number
```
mjtalbot, line 31: 🤦

app/src/main/java/app/rive/runtime/example/ViewPagerActivity.kt
```diff
@@ -0,0 +1,85 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.ListAdapter
+import androidx.recyclerview.widget.RecyclerView
+import app.rive.runtime.example.databinding.ViewPagerRiveWrapperBinding
+import app.rive.runtime.example.databinding.ViewPagerBinding
+
+
+class ViewPagerActivity : AppCompatActivity() {
+    private lateinit var binding: ViewPagerBinding
```
mjtalbot, line 17: ohh i guess this does!

cpp/src/bindings/bindings_renderer_skia.cpp
```diff
@@ -49,9 +49,10 @@ extern "C"
     JNIEXPORT void JNICALL
     Java_app_rive_runtime_kotlin_renderers_RendererSkia_cppStart(JNIEnv*,
                                                                  jobject,
-                                                                 jlong rendererRef)
+                                                                 jlong rendererRef,
+                                                                 jlong frameTimeNs)
     {
-        reinterpret_cast<JNIRendererSkia*>(rendererRef)->start();
+        reinterpret_cast<JNIRendererSkia*>(rendererRef)->start(frameTimeNs);
```
mjtalbot, line 55: oh is this to allow animations to kinda start in the middle more? this is pretty cool. its not (always) going to work with statemachines that can have user interaction of course

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
```diff
@@ -432,6 +432,19 @@ open class RiveAnimationView(context: Context, attrs: AttributeSet? = null) :
         alignment: Alignment = Alignment.CENTER,
         loop: Loop = Loop.AUTO,
     ) {
+        if (resId == rendererAttributes.resourceId && renderer.file != null) {
```
mjtalbot, line 435: what is the significance of this, does this mean we can skip a few things when re initializing a RiveAnimationView?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -539,10 +550,11 @@ open class RiveArtboardRenderer(
 
     private fun setArtboard(artboard: Artboard) {
         // clean up any previous artboard if one was set
-        this.activeArtboard?.dispose()
-        this.activeArtboard = artboard
-
-
+        if (artboard != this.activeArtboard) {
```
mjtalbot, line 553: i'm guessing this is something we need to do to deal with rehydrating?

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
```diff
@@ -539,10 +550,11 @@ open class RiveArtboardRenderer(
...
+            this.activeArtboard?.dispose()
```
mjtalbot, line 554: do we get a situation where we dispose things, athat are already disposed as part of "disposeDependencies"? or can things be disposed of multiple times

kotlin/src/main/java/app/rive/runtime/kotlin/RiveTextureView.kt
```diff
@@ -89,7 +90,7 @@ abstract class RiveTextureView(context: Context, attrs: AttributeSet? = null) :
 
     @CallSuper
     override fun onSurfaceTextureDestroyed(surface: SurfaceTexture): Boolean {
-        // Returning true will `release()` for us
-        return true
+        viewSurface.release()
+        return false
```
mjtalbot, line 94: what does this change fix?

app/src/main/java/app/rive/runtime/example/RecyclerActivity.kt
umberto-sonnino: 🫣

app/src/main/java/app/rive/runtime/example/ViewPagerActivity.kt
umberto-sonnino: Yeah! Got this from the repro we got, very Android-y way of doing things, not sure I like this better than the old version

cpp/src/bindings/bindings_renderer_skia.cpp
umberto-sonnino: Not exactly! I found some issues when restarting some views that had to do with timing. We have a ns-timer in C++ that handles some of the starting up of the pacing, but we were later using the time passed in by the `doFrame()` callback from the Android `Choreographer`. Unfortunately, these times don't line up and they tend to cause issues with `elapsed`, so I went back to our original logic and started using `System.nanoTime` everywhere - that way we have a single source of truth

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
umberto-sonnino: This was a pretty important fix wrt `RecyclerViews` and `ViewPagers`!
In fact, when you reuse this `View` gets properly recycled, it has all the values from the *old* view. And that means that, sometimes, we might be re-using the same View for a different resource. 

For example:
I now changed our `RecyclerView` demo to be alternating between the `basketball` and the `circle_move` animation (before it was always using the same animation). While scrolling, the list would reuse the `basketball` _view_ to initialize a `circle_move` instead, but it had all the old properties cached! This clearly led to our `View` failing in that scenario as it was trying to start up the animation from the other `file` 

This fix also impacts another piece, that is we were re-initializing `File`s all the time, even when that same `File` had already been set! With this and the `setRiveFile` fix below, this is now a bit more efficient in just init'ing what we already have instead.

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
umberto-sonnino: Yeah, same as for the `File`, in this case we're just trying to be smarter about initializing resources only when we need. If everything's already set up, we can definitely skip this step, which fits very well in the case of scrolling views..!

kotlin/src/main/java/app/rive/runtime/kotlin/RiveArtboardRenderer.kt
umberto-sonnino: This should *never* happen! If it does, we'll throw as we're effectively trying to delete the underlying C++ object twice

kotlin/src/main/java/app/rive/runtime/kotlin/RiveTextureView.kt
umberto-sonnino: This was another interesting one, probably the first thing that I fixed. Following the docs, my understanding was that by returning `true` here, we'd be leaving the cleanup up to the OS. 
But then I turned on some warnings with this bit:
```kotlin
try {
 Class.forName("dalvik.system.CloseGuard")
   .getMethod("setEnabled", Boolean::class.javaPrimitiveType)
   .invoke(null, true)
} catch (e: ReflectiveOperationException) {
 throw RuntimeException(e)
}
``` 

And I started seeing warnings `W/System: A resource failed to call release.`
After trying a few things, this made the warning go away.

kotlin/src/main/java/app/rive/runtime/kotlin/RiveAnimationView.kt
mjtalbot: wait so we have 

setRiveResource
setRiveBytes
setRiveFile

setRiveBytes, actually calls setRiveFile. 
and setRiveResource calls setRiveBytes...

but now we are changing it so that  we can bypass setting the bytes from the resource stream, and instead go straight to setRiveFile, with what we've already had set on the renderer?

PR #296, opened 2023-06-24
Fix leaking issue when add lifecycle observer

hoangthan:
Since the context is using ActivityLifeCycle. So trying to add lifeCycleObserver without remove when the view is being detached will causing leak memory issue.

### DIFF
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/RiveTextureView.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/RiveTextureView.kt
@@ -44,11 +44,6 @@ abstract class RiveTextureView(context: Context, attrs: AttributeSet? = null) :
         (sInNS / refreshRateHz).toLong()
     }
 
-    init {
-        // Attach the observer to give us lifecycle hooks.
-        getContextAsType<LifecycleOwner>()?.lifecycle?.addObserver(lifecycleObserver)
-    }
-
     private inline fun <reified T> getContextAsType(): T? {
         var ctx = context
         while (ctx is ContextWrapper) {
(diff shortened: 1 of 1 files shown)


umberto-sonnino:
> Since the context is using ActivityLifeCycle. So trying to add lifeCycleObserver without remove when the view is being detached will causing leak memory issue.
> 
> The lifeCycleObserver is already remove itself once the lifeCycle is reached `onDestroyed`. But with activity, it will be never called until the activity is going on destroyed.

The behavior for this flow has now been revised and adjusted by re-evaluating the `LifecycleOwner` on attach. Do you still the leak on the latest version?

stevezuju:
> > Since the context is using ActivityLifeCycle. So trying to add lifeCycleObserver without remove when the view is being detached will causing leak memory issue.
> > The lifeCycleObserver is already remove itself once the lifeCycle is reached `onDestroyed`. But with activity, it will be never called until the activity is going on destroyed.
> 
> The behavior for this flow has now been revised and adjusted by re-evaluating the `LifecycleOwner` on attach. Do you still the leak on the latest version?

Is `5.1.5` the latest version ?

stevezuju:
> > Since the context is using ActivityLifeCycle. So trying to add lifeCycleObserver without remove when the view is being detached will causing leak memory issue.
> > The lifeCycleObserver is already remove itself once the lifeCycle is reached `onDestroyed`. But with activity, it will be never called until the activity is going on destroyed.
> 
> The behavior for this flow has now been revised and adjusted by re-evaluating the `LifecycleOwner` on attach. Do you still the leak on the latest version?

The life-cycler you're using is activity lifecycle. The observer will be never removed in case activity not going on destroyed.
You should remove the observer while the RiveTextureView is going to detached.

umberto-sonnino:
> > > Since the context is using ActivityLifeCycle. So trying to add lifeCycleObserver without remove when the view is being detached will causing leak memory issue.
> > > The lifeCycleObserver is already remove itself once the lifeCycle is reached `onDestroyed`. But with activity, it will be never called until the activity is going on destroyed.
> > 
> > 
> > The behavior for this flow has now been revised and adjusted by re-evaluating the `LifecycleOwner` on attach. Do you still the leak on the latest version?
> 
> The life-cycler you're using is activity lifecycle. The observer will be never removed in case activity not going on destroyed. You should remove the observer while the RiveTextureView is going to detached.

The latest version is `6.0.0`, is that what you're using?
Changes were introduced [here](https://github.com/rive-app/rive-android/commit/9e2beb8405b8b370ff8d24c08895178efd8ff100) but we're not referencing the Activity directly.

PR #331, opened 2024-06-20
Fix bug crash on Android 15 (with 16KB page size support)

vegeta2102:
### Description 

Following on this issue https://github.com/rive-app/rive-android/issues/329 and the guidance from AndroidDeveloper
https://developer.android.com/guide/practices/page-sizes#compile-r26-lower
native-code has to rebuilt in order to be compatible with Android 15.

```
target_link_options(${CMAKE_PROJECT_NAME} PRIVATE "-Wl,-z,max-page-size=16384")
```
Updated CMakeLists.txt to enable 16 KB ELF alignment.

### DIFF
--- a/kotlin/src/main/cpp/CMakeLists.txt
+++ b/kotlin/src/main/cpp/CMakeLists.txt
@@ -128,3 +128,5 @@ target_link_libraries(rive-android
         ${gles-lib}
         jnigraphics
         )
+
+target_link_options(rive-android PRIVATE "-Wl,-z,max-page-size=16384")


vegeta2102:
@umberto-sonnino 
Can you have a look at it?

georgmay:
Сan somebody please look into this issue?

ErikUggeldahl:
Hi all. I'm the Android DevRel for Rive and am actively looking into this issue. The solution posed in this PR may not be sufficient, as it only affects our C++ files but not the associated `libc++_shared.so`. So we're now investigating bumping the NDK version which would ensure everything is 16kb aligned. Stay tuned for more as we test this solution more.

ErikUggeldahl:
Hi all. This thread slipped my mind but this was fixed as of 9.13.0. We ultimately went with NDK r27, [as described here](https://developer.android.com/guide/practices/page-sizes#compile-r27). Closing as this is now resolved.

Issue #341, opened 2024-09-20
App crash due to missing libraries for `x86`, `armeabi-v7a`, `armeabi`


shakil-bin-karim:
### Description

I'm encountering a crash in my Android app for some phones when using the Rive Android library. The app crashes on startup with the following error:

`androidx.startup.StartupException: com.getkeepsafe.relinker.MissingLibraryException: Could not find 'librive-android.so'. Looked for: [x86, armeabi-v7a, armeabi], but only found: [arm64-v8a].`

This indicates that the librive-android.so native library is missing for certain architectures, specifically x86, armeabi-v7a, and armeabi. The app only includes the library for the arm64-v8a architecture.

### Provide a Repro

The crash occurs when running the app on devices with architectures other than arm64-v8a, such as armeabi-v7a or x86.

- If the problem is small enough, paste a code snippet
- Otherwise, you can attach a ZIP archive or paste a link to a GitHub repository with the app bundle
  that is reproducing the issue

Steps to reproduce the issue:

1. Integrate the Rive Android library into your Android app.
2. Build and run the app on device with an armeabi-v7a or x86 architecture.
3. Observe that the app crashes on startup with the missing library error.

#### Source `.riv`/`.rev` file

N/A – The issue occurs during app initialization, before any .riv files are loaded.

### Expected behavior

The app should start without crashing, and the librive-android.so library should be correctly loaded for all supported architectures (armeabi-v7a, arm64-v8a, x86, etc.).

#### Screenshots

N/A

### Device & Versions (please complete the following information)

- Device: Various devices with devices with no `arm64-v8a` architecture from manufactures like ZTE, Oppo, Older Samsung devices, etc.
- SDK Level: Android SDK API Level 35

#### Additional context

Here are the full stack traces from Firebase Crashlytics:

```
Fatal Exception: java.lang.RuntimeException: Unable to get provider androidx.startup.InitializationProvider: androidx.startup.StartupException: com.getkeepsafe.relinker.MissingLibraryException: Could not find 'librive-android.so'. Looked for: [x86, armeabi-v7a, armeabi], but only found: [arm64-v8a].
       at android.app.ActivityThread.installProvider(ActivityThread.java:5927)
       at android.app.ActivityThread.installContentProviders(ActivityThread.java:5516)
       at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5455)
       at android.app.ActivityThread.-wrap2(ActivityThread.java)
[...]

PR #355, opened 2025-02-15
fix: native library loading

DatL4g:
Introduce a NativeLoader which takes care of loading native libraries. ReLinker is not needed on API Level > 23 and causes more trouble than it solves.

Refs: #341

### DIFF
--- a/kotlin/src/main/java/app/rive/runtime/kotlin/RiveInitializer.kt
+++ b/kotlin/src/main/java/app/rive/runtime/kotlin/RiveInitializer.kt
@@ -3,6 +3,7 @@ package app.rive.runtime.kotlin
 import android.content.Context
 import androidx.startup.Initializer
 import app.rive.runtime.kotlin.core.Rive
+import app.rive.runtime.kotlin.core.RiveInitialized
 
 /**
  * Initializes Rive; needs to be done at startup.
@@ -37,9 +38,9 @@ import app.rive.runtime.kotlin.core.Rive
  * In fact, if you want to provide a custom renderer type you'll need to init Rive manually.
  */
 
-class RiveInitializer : Initializer<Unit> {
-    override fun create(context: Context) {
-        return Rive.init(context)
+class RiveInitializer : Initializer<RiveInitialized> {
+    override fun create(context: Context): RiveInitialized {
+        return RiveInitialized(Rive.init(context))
     }
 
     override fun dependencies(): List<Class<out Initializer<*>>> {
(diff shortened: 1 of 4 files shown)


ernestoyaquello:
Does this fix #341, or is it just related to it? I am not entirely sure what's going on here. 😅

DatL4g:
@ernestoyaquello It's related to the mentioned issue since it's a known bug in ReLinker for years.
It will probably fix the issue as well, at least I don't see any reason why it should not work after merging.  
I also use this kind of method in production apps and works without crashes, no matter if real/emulator or architecture.

## More Context

Rive uses ReLinker to load native code which is only needed on older Android API Levels (below 23, Android Marshmallow).  
It uses it's own native code loading process instead of `System.loadLibrary` because older phones did not support this properly.

Starting with Android 23 and up the `System.loadLibrary` works without issues.

## Code Explanation

The updated and introduced code checks if the phone is old, then it uses ReLinker, otherwise it uses the normal `System.loadLibrary` method.  
Additionally it won't crash the application if Rive could not be initialized, the `Rive.init` method returns a `Boolean` whether it could be loaded.
This way people can dynamically disable animations, only thing to mention is that it will always return `false` if it's called more than once.

## Work Around while not merged

I use this code to initialize rive properly:  
Btw I am using a more complex version of [`NativeLoader`](https://github.com/DatL4g/Sekret/blob/master/sekret-lib/src/androidMain/kotlin/dev/datlag/sekret/NativeLoader.kt) here which is not required for rive.

```kotlin
/**
 * Retrieves the rive native binary name using reflection, falls back to "rive-android"
 * Uses NativeLoader to load the native binary
 * Sets the default renderer using reflection
 */
fun Rive.initSafely(
    context: Context,
    defaultRenderer: RendererType = defaultRendererType
) {
    val riveClass = "app.rive.runtime.kotlin.core.Rive"
    val libName = runCatching {
        val clazz = Class.forName(riveClass)
        val field = clazz.getDeclaredField("RIVE_ANDROID")
        field.isAccessible = true
        field.get(null) as? String
    }.getOrNull()?.trim()?.ifBlank { null } ?: "rive-android"

    val libLoaded = NativeLoader.loadLibrary(context, libName)
    val rendererSet = this.defaultRendererType == defaultRenderer || runCatching {
        val clazz = Class.forName(riveClass)
        val field = clazz.getDeclaredField("defaultRendererType")
        field.isAccessible = true
        field.set(clazz, defaultRenderer)
    }.isSuccess

    if (libLoaded && rendererSet) {
        initializeCppEnvironment()
    } else {
        init(context, defaultRenderer)
    }
}
```

umberto-sonnino:
HI @DatL4g, thanks for your contribution! 

It looks very promising and we're keen to find a way to resolve #341 for all those who reported it. As I mentioned in the issue thread, before integrating this PR, we'd like to test this and make sure this is solving the crashes.

Could you give us more specific repro steps? Which specific devices/ABIs have you confirmed experience the crash with the current implementation?

Once we validate this, we're happy to merge it

Issue #381, opened 2025-06-22
ANR: condition_variable::wait (from ~JNIRenderer)


leemhenson:
### Description

We're seeing this error reported by Android Vitals. Stack trace is:

```
"main" tid=1 Native
  #00  pc 0x000000000004f55c  /apex/com.android.runtime/lib64/bionic/libc.so (syscall+28)
  #01  pc 0x0000000000054110  /apex/com.android.runtime/lib64/bionic/libc.so (__futex_wait_ex+144)
  #02  pc 0x00000000000c0ad8  /apex/com.android.runtime/lib64/bionic/libc.so (pthread_cond_wait+72)
  #03  pc 0x00000000000df3b8  /data/app/~~ijAZxYx3Bxh1YVoiATSx5w==/io.blueheart.socialanxiety-UGK7laDihT7mWlMGFAijEQ==/split_config.arm64_v8a.apk (std::__ndk1::condition_variable::wait(std::__ndk1::unique_lock<std::__ndk1::mutex>&)+24) (BuildId: 6783a7f3a0d9c67c55a74cadc441ed55aaa493da)
  #04  pc 0x000000000017e2cc  /data/app/~~ijAZxYx3Bxh1YVoiATSx5w==/io.blueheart.socialanxiety-UGK7laDihT7mWlMGFAijEQ==/split_config.arm64_v8a.apk (BuildId: 2f494393cfaaee8d79090be125ccb7f8536b17cc)
  #05  pc 0x0000000000183cf8  /data/app/~~ijAZxYx3Bxh1YVoiATSx5w==/io.blueheart.socialanxiety-UGK7laDihT7mWlMGFAijEQ==/split_config.arm64_v8a.apk (rive_android::WorkerThread::runAndWait+136) (BuildId: 2f494393cfaaee8d79090be125ccb7f8536b17cc)
  #06  pc 0x00000000001896c8  /data/app/~~ijAZxYx3Bxh1YVoiATSx5w==/io.blueheart.socialanxiety-UGK7laDihT7mWlMGFAijEQ==/split_config.arm64_v8a.apk (rive_android::JNIRenderer::~JNIRenderer+48) (BuildId: 2f494393cfaaee8d79090be125ccb7f8536b17cc)
[...]

PR #394, opened 2025-08-03
Make WorkerThread re-entrant & ~JNIRenderer async to address #381

cojo:
Spent some time on this this weekend as the relevant issue (#381) is completely blocking our ability to upgrade to Expo 53 / newArch which is a blocking issue for our company and our product for other reasons.

My goal was to make the minimal changeset necessary to address the destructor ANR while also maximizing chances of avoiding any other similar or related deadlocks in the relevant file(s).

I am able to reproduce this ANR ~90% of the time on my Pixel 7 device on both debug and release builds using the rive-react-native SDK version 9.4.0 - the reproduction is simply to mount a Rive component with autoplay enabled, then unmount it.

I have built locally and confirmed that this changeset fixes the issue completely (given the high repro rate before and no repro since, I am very confident it is resolved now) and also has introduced no other issues with Rive in our product (which uses it heavily).

The basic idea here is that, as the constructor itself already implies / intends based on its comment:
```
    // Grab a Global Ref to prevent Garbage Collection to clean up the object
    //  from under us since the destructor will be called from the render thread
    //  rather than the UI thread.
```

The destructor should be called from something _other than the UI thread itself_.  However, right now, in practice, at least from the rive-react-native usage pattern, this is _exactly_ the opposite of what happens.

So this change ensures that _any_ JNI caller will now queue up the destructor onto the worker thread instead of the calling thread (which might very well be the UI thread).

In order to facilitate this, it also adds some additional protections to ensure the worker is properly re-entrancy-safe.  It additionally makes some minor semantic changes to some other lock / mutex usages which i do not believe were 100% correct previously.

Happy to answer any questions or address any feedback on this; I'm not sure how quickly the project typically adopts fixes like this, but if you can give context on that it would be helpful, as we will likely need to make our own local build and deployment of our fork if that process will take longer than a few days (we urgently need to get this build out and this is the one thing blocking the release).

Thanks in advance!
-cojo

### DIFF
--- a/kotlin/src/main/cpp/include/helpers/worker_thread.hpp
+++ b/kotlin/src/main/cpp/include/helpers/worker_thread.hpp
@@ -83,14 +83,23 @@ public:
         {
             return; // Early out that doesn't require a mutex!
         }
-        std::lock_guard<std::mutex> threadLock(mWorkMutex);
-        while (m_lastCompletedWorkID < workID)
-        {
-            m_workedCompletedCondition.wait(mWorkMutex);
-        }
+        std::unique_lock<std::mutex> lk(mWorkMutex);
+        m_workedCompletedCondition.wait(lk, [this, workID] {
+            return m_lastCompletedWorkID >= workID;
+        });
     }
 
-    void runAndWait(Work&& work) { waitUntilComplete(run(std::move(work))); }
+    void runAndWait(Work&& work) {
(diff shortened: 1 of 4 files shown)


ErikUggeldahl:
Thanks again for your help with this @cojo. Glad we were able to get to the bottom of this.

PR #417, opened 2025-10-17
Implement modern responsive layouts file instead of old file

tylernij:
* Replace Layouts file with [new one](https://github.com/rive-app/rive-flutter/blob/master/example/lib/examples/responsive_layouts.dart)
* Fix Edge-to-Edge hiding scrollbar behind top bar


tylernij:
I'm going to close this PR and re-do it from clean slate

PR #418, opened 2025-10-17
layouts: Replace layout test with responsive layouts

tylernij:
1. Fix compose activity being hidden due to E2E
2. Use default fit type for responsive layout
3. Fix animations not playing
4. Remove no longer needed riv file


lancesnider:
lgtm! I'll leave Erik to approve.

HayesGordon (changes requested):
First Android PR, exciting!

I'm just blocking this from merging as the changes need to happen on the Mono repository, that then get's automatically pushed downstream. If we merge anything directly downstream we'll be in a state where the mono repo is out of date and we'd have to manually fix things.

Issue #359, opened 2025-03-13
RiveException: Accessing disposed C++ object Artboard.


thorbenprimke:
<-- Thanks for using Rive!

If something isn't working like you expected, please:
The library's code is causing a crash. 

-->

### Description
It looks like during a cleanup in the Rive code, a native exception is triggered. 

Also this exception has the entire stacktrace in the message rather than the stacktrace itself. This caused Instabug to not display a stacktrace.

```
java.lang.Throwable: 
app.rive.runtime.kotlin.core.errors.RiveException: Accessing disposed C++ object Artboard. 
    Dispose_Trace.Start(Unknown Source)
    app.rive.runtime.kotlin.core.NativeObject.dispose(Unknown Source:14)
    app.rive.runtime.kotlin.core.NativeObject.release(Unknown Source:15)
    app.rive.runtime.kotlin.core.NativeObject.dispose(Unknown Source:41)
    app.rive.runtime.kotlin.core.NativeObject.release(Unknown Source:15)
    app.rive.runtime.kotlin.core.File.release(Unknown Source:3)
    app.rive.runtime.kotlin.controllers.RiveFileController.setFile(Unknown Source:28)
    app.rive.runtime.kotlin.RiveAnimationView.setRiveResource$lambda$11(Unknown Source:2)
    app.rive.runtime.kotlin.RiveAnimationView.g(Unknown Source:0)
    c4.a.invoke(Unknown Source:19)
    app.rive.runtime.kotlin.RiveAnimationView.loadFileFromResource(Unknown Source:152)
    app.rive.runtime.kotlin.RiveAnimationView.setRiveResource(Unknown Source:42)
    app.rive.runtime.kotlin.RiveAnimationView.setRiveResource$default(Unknown Source:96)
    p0.i.invoke(Unknown Source:2612)
    x3.k1.b(Unknown Source:79)
    mn.L0.c(Unknown Source:197)
    mn.K0.invokeSuspend(Unknown Source:12)
    Tl.a.resumeWith(Unknown Source:5)
    jn.J.run(Unknown Source:121)
    jn.W.F0(Unknown Source:23)
    jn.h.p(Unknown Source:78)
    jn.h.C(Unknown Source:32)
    jn.h.resumeWith(Unknown Source:16)
    mn.L0.n(Unknown Source:103)
    mn.L0.l(Unknown Source:8)
    Hf.k.d(Unknown Source:13)
    Hf.k.n(Unknown Source:31)
[...]

PR #429, opened 2025-12-08
[#424] Fix: app.rive.runtime.kotlin.core.errors.RiveException: Accessing disposed C++ object RiveArtboardRenderer.

### LINKED ISSUE #424: Crash! app.rive.runtime.kotlin.core.errors.RiveException: Accessing disposed C++ object RiveArtboardRenderer.
Steps to reproduce:
LazyColumn inside each row  button with rive.
Quite fast scroll up/down for few seconds.

And we have get a this crash. We are trying to stop animation by onRelease.  `listItem.onRelease = { }` do dispose , don't do dispose but still get this crash.

```
AndroidRuntime                  E  FATAL EXCEPTION: Thread-15
                                   Process: com.letyco, PID: 27929
                                   java.lang.Throwable: 
                                   app.rive.runtime.kotlin.core.errors.RiveException: Accessing disposed C++ object RiveArtboardRenderer.
                                       app.rive.runtime.kotlin.core.NativeObject.getCppPointer(NativeObject.kt:55)
                                       app.rive.runtime.kotlin.renderers.Renderer.getWidth(Renderer.kt:215)
                                       app.rive.runtime.kotlin.renderers.RiveArtboardRenderer.resizeArtboard(RiveArtboardRenderer.kt:36)
                                       app.rive.runtime.kotlin.renderers.RiveArtboardRenderer.draw(RiveArtboardRenderer.kt:51)
InputDispatcher                 E  channel 'b67e987 com.letyco/com.letyco.MainActivity' ~ Channel is unrecoverably broken and will be disposed!
```

dabrosch:
[#424, #359] Fix: app.rive.runtime.kotlin.core.errors.RiveException: Accessing disposed C++ object RiveArtboardRenderer.

The crash is a TOCTOU race on the render worker: `draw()` can call `resizeArtboard()` (which reads renderer width/height) while another thread is deleting the renderer, so the C++ pointer is disposed between the `hasCppObject` check and the dimension read.

Master partially addressed this in #11831 by splitting `resizeArtboard()` into separate frameLock/fileLock sections, but resize still runs **outside** `draw()`'s `frameLock`. `delete()` can still win the race once `requireArtboardResize` is cleared and `resizeArtboard()` begins.

This PR keeps resize and draw under the same lock scope: acquire `frameLock`, then `fileLock` (matching `controller.advance()`), run resize if needed, then draw. `resizeArtboard()` is private and assumes its caller already holds both locks. The `disposeDependencies()` fileLock override is removed; with the new ordering, it would invert lock acquisition relative to `draw()` and deadlock.

`RiveFileController.isActive` is marked `@Volatile` so the worker thread sees UI-thread deactivation promptly inside `draw()`.

### Tests
The old dispose-during-draw tests used `RiveFileController` without a `File`, so `draw()` and `activeArtboard` fell back to synchronizing on **different** objects. They passed even when `fileLock` was removed from `draw()`. Tests now load `R.raw.empty` so both paths share `File.fileLock`, matching production `RiveAnimationView` usage.
- `deleteRendererDuringFrame` — `frameLock` only (no File)
- `drawHoldsFileLockBlocksActiveArtboardMutation` — UI `activeArtboard` mutation blocks on `fileLock`
- `disposeArtboardDuringFrameAfterEnteringSyncBlock` / `BeforeEnteringSyncBlock` — artboard disposal while draw holds `fileLock`
- `deleteRendererBetweenCppObjectCheckAndDimensionRead_doesNotCrash` — #424 regression (delete during resize pause)

Lock-order validation: removing `fileLock` from `draw()` fails the file-lock tests; removing `frameLock` fails the frame-lock / #424 tests.

Issues fixed: https://github.com/rive-app/rive-android/issues/424 and https://github.com/rive-app/rive-android/issues/359


kotlin/src/main/java/app/rive/runtime/kotlin/renderers/RiveArtboardRenderer.kt
```diff
@@ -92,7 +101,10 @@ open class RiveArtboardRenderer(
     }
 
     override fun disposeDependencies() {
-        // Lock to make sure things are disposed in an orderly manner.
-        synchronized(controller.file?.lock ?: this) { super.disposeDependencies() }
+        // Lock to ensure dependencies are disposed in an orderly manner and to coordinate
+        // with any ongoing file/artboard operations that might be using these dependencies.
+        synchronized(controller.file?.lock ?: this) {
+            super.disposeDependencies()
+        }
```
dabrosch: I just realized this could cause a deadlock with the above code. I am not even fully convinced that we really do need a File Lock on this, but if we do keep it, the order of lock obtainment needs to match the order in draw, so Frame Lock then File Lock.

RostyslavProkopenko:
bump

dabrosch:
Merged with main again, but noticed that the changes pushed since last time still did not actually fully fix the deadlock issue, so this PR is still needed (I just updated the PR description again).

dabrosch:
This PR also would seem to address https://github.com/rive-app/rive-android/issues/359

I also noticed that we are doing this in RiveFileController:
```
val mArtboard = this.activeArtboard ?: return null
        synchronized(mFile.fileLock) 
```

But in RiveArtboardRenderer we do:
`synchronized(controller.file?.fileLock ?: this)`

So theoretically we could end up with the renderer and controller locking on different objects, which obviates the purpose of it; while in practice it doesn't look likely to happen.

PR #447, opened 2026-04-10
Fix Fit.LAYOUT artboard size when onMeasure is deferred or state machine animates artboard width

### LINKED ISSUE #446: Fit.LAYOUT artboard size incorrect when onMeasure is deferred or state machine animates artboard width
### Description

`Fit.LAYOUT` artboard size is incorrect under two conditions:

1. **`layoutScaleFactorAutomatic` defaults to `1.0`** and is only set to device density in `onMeasure()`. Since the view hasn't been measured yet when `setRiveFile()` is called, `resizeArtboard()` computes `pixelWidth / 1.0` instead of `pixelWidth / density` — the artboard ends up density× too wide. iOS avoids this by [initializing the scale factor at view init time](https://github.com/rive-app/rive-ios/blob/68513e97/Source/RiveView.swift#L38-L44). Originally reported in [rive-nitro-react-native#206](https://github.com/rive-app/rive-nitro-react-native/issues/206), simple workaround [possible](https://github.com/rive-app/rive-nitro-react-native/pull/209) by setting `layoutScaleFactor` to density before `setRiveFile`.

2. **When a state machine animates the artboard width**: keyframes overwrite `m_Width` every frame via `CoreRegistry::setDouble(artboard, widthPropertyKey, ...)` during `advance()`. Since `resizeArtboard()` only runs once (one-shot `requireArtboardResize` flag), the artboard stays at the keyframe value.

For reference, the iOS runtime approaches this differently: it [initializes scale factor at view init time](https://github.com/rive-app/rive-ios/blob/68513e97/Source/RiveView.swift#L38-L44) and [re-applies artboard dimensions every frame in `drawRive`](https://github.com/rive-app/rive-ios/blob/68513e97/Source/RiveView.swift#L446-L452) rather than using a one-shot flag. It also re-applies on size change in [`drawableSizeDidChange`](https://github.com/rive-app/rive-ios/blob/68513e97/Source/RiveView.swift#L490-L501). Not saying this is necessarily the right approach for Android, but it avoids both issues above.

### Versions, Device, and Other Information

- Rive Android Runtime Version: 11.3.2
- Rive API: Legacy (View-based)
- Device: Android API 35 emulator (1080x2220, 440dpi), also reported on OnePlus CPH2719, Samsung Galaxy Fold 3
- App Minimum SDK Level: 21
- App Target SDK Level: 36
- Frequency: Issue 1 — 100% when setRiveFile is called before onMeasure. Issue 2 — 9/10 with layout_test.riv.

### Reproduction Steps

**Issue 1** (scale factor default): Any `.riv` with `Fit.LAYOUT`. The view hasn't been measured when `setRiveFile` is called, so `layoutScaleFactorAutomatic` is still `1.0`. Confirmed via logcat:
```
configure: reload=true fit=LAYOUT viewSize=0x0px
configure post-setRiveFile: artboardW=1920.0 artboardH=1080.0
onLayout: artboardW=1920.0 artboardH=1080.0  ← still intrinsic size after layout
```

Reproduced in bare Android by suppressing `onMeasure`:
```
✗ Artboard 1080 dp ≠ View 393 dp (2.8x too wide!)
```
The 2.75x factor matches exactly the device density — `1080px / 1.0` instead of `1080px / 2.75`.

**Issue 2** (keyframe overwrite): `layout_test.riv` with `Fit.LAYOUT` — the state machine animates the artboard width property, overwriting the value set by `resizeArtboard()` every frame. Reproduced 9/10 launches.

### Source `.riv`/`.rev` File

- Issue 1: Any `.riv` with `Fit.LAYOUT` (tested with `GradientBorder.riv` from [user's repro](https://github.com/Nezz/expo-repro/tree/repro/RiveSize))
- Issue 2: `layout_test.riv` from rive-android example app

### Expected Behavior

Artboard dimensions should match `viewWidth / density` for `Fit.LAYOUT`.

### Additional context

For reference, iOS handles this in:
- [`RiveView.swift L38-44`](https://github.com/rive-app/rive-ios/blob/68513e97/Source/RiveView.swift#L38-L44) — scale factor initialized at view init
- [`drawRive` L446-452](https://github.com/rive-app/rive-ios/blob/68513e97/Source/RiveView.swift#L446-L452) — artboard dimensions re-applied every frame
- [`drawableSizeDidChange` L490-501](https://github.com/rive-app/rive-ios/blob/68513e97/Source/RiveView.swift#L490-L501) — re-applied on size change

mfazekas:
Fixes #446

## Problem

See #446 for full details. Two conditions where `Fit.LAYOUT` artboard size is wrong:

1. `layoutScaleFactorAutomatic` defaults to `1.0`, only set to device density in `onMeasure()`. The view hasn't been measured yet when `setRiveFile()` is called, so `resizeArtboard()` uses the wrong scale factor.
2. State machine keyframes can overwrite `m_Width` every frame during `advance()`. The one-shot `requireArtboardResize` flag is already consumed, so `resizeArtboard()` never corrects it.

For reference, iOS re-applies artboard dimensions [every frame in `drawRive`](https://github.com/rive-app/rive-ios/blob/68513e97/Source/RiveView.swift#L446-L452) rather than using a one-shot flag.

## Changes

1. **`RiveAnimationView` init** — set `layoutScaleFactorAutomatic` to device density at construction time, like [iOS does](https://github.com/rive-app/rive-ios/blob/68513e97/Source/RiveView.swift#L38-L44), instead of leaving it at `1.0` until `onMeasure`
2. **`RiveArtboardRenderer.draw()`** — call `resizeArtboard()` every frame when `fit == Fit.LAYOUT`
3. **`RiveArtboardRenderer.resizeArtboard()`** — re-arm flag when artboard is null or surface is 0×0
4. **`RiveAnimationView.onSurfaceTextureSizeChanged/Available()`** — set `requireArtboardResize = true`

## Verified

- Issue 1: confirmed via logcat (`artboardW=1920` stuck at intrinsic size). Bare Android reproducer simulates by suppressing `onMeasure`.
- Issue 2: `layout_test.riv` — 9/10 wrong before fix, 10/10 correct after.

### DIFF
--- a/app/src/androidTest/kotlin/app/rive/runtime/example/FitLayoutReproTest.kt
+++ b/app/src/androidTest/kotlin/app/rive/runtime/example/FitLayoutReproTest.kt
@@ -1,0 +1,175 @@
+package app.rive.runtime.example
+
+import android.widget.FrameLayout
+import androidx.test.core.app.ActivityScenario
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import app.rive.runtime.kotlin.RiveAnimationView
+import app.rive.runtime.kotlin.core.File
+import app.rive.runtime.kotlin.core.Fit
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import kotlin.time.Duration.Companion.milliseconds
+
+/**
+ * Reproducer for the Fit.LAYOUT race condition:
+ *
+ * When setRiveFile(fit = Fit.LAYOUT) is called on a 0×0 view (before the
+ * first measure/layout pass), the artboard sometimes stays at its intrinsic
+ * size instead of resizing to match the view.
+ *
+ * This happens because setupScene() nulls activeArtboard then sets
+ * requireArtboardResize=true, and the render thread can consume that flag
+ * while activeArtboard is still null.
+ */
+@RunWith(AndroidJUnit4::class)
+class FitLayoutReproTest {
+
+    /**
+     * Programmatically add a RiveAnimationView and call setRiveFile with
+     * Fit.LAYOUT before the view has been measured. Verify the artboard
+     * resizes to the view size (not stuck at intrinsic size).
+     */
+    @Test
+    fun fitLayoutResizesWhenSetBeforeMeasure() {
+        val activityScenario = ActivityScenario.launch(EmptyActivity::class.java)
+        lateinit var riveView: RiveAnimationView
+        val laidOutLatch = CountDownLatch(1)
+
+        val viewWidthPx = 800
+        val viewHeightPx = 400
+
+        activityScenario.onActivity { activity ->
+            val riveBytes = activity.resources
+                .openRawResource(R.raw.layout_test)
+                .readBytes()
+            val riveFile = File(riveBytes)
+
+            riveView = RiveAnimationView(activity)
+            riveView.layoutParams = FrameLayout.LayoutParams(viewWidthPx, viewHeightPx)
+
+            // Add to container — triggers measure/layout asynchronously
+            activity.container.addView(riveView)
+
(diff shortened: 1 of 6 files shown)


mfazekas:
Split into two independent PRs: #449 (scale factor default) and #450 (per-frame resize).

PR #451, opened 2026-04-13
fix: use emulated TLS on armeabi-v7a to fix linker crashes on API < 29

### LINKED ISSUE #445: Android startup crash in 11.3.1 on armeabi-v7a / older devices: UnsatisfiedLinkError "unknown reloc   type 17"
## Bug report

  After upgrading `app.rive:rive-android` from `10.4.4` to `11.3.1`, our Android app started crashing
  on startup on a subset of devices, mostly older / 32-bit Android devices.

  The crash happens during `Rive.init(...)`, before any screen is shown.

  ## Versions

  - Previous working version: `10.4.4`
  - Crashing version: `11.3.1`

  ## Crash

  ```text
  java.lang.UnsatisfiedLinkError: dlopen failed: unknown reloc type 17 @ 0x... (11815)
```
  App code:
```
  Rive.init(this, defaultRenderer = RendererType.Rive)
```
  Relevant stacktrace:
```
  java.lang.UnsatisfiedLinkError: dlopen failed: unknown reloc type 17 @ 0xbe8f93ac (11815)
      at <app>.Application.initializeRive(Application.kt:30)
      at <app>.Application.onCreate(Application.kt:43)
      at app.rive.runtime.kotlin.core.Rive.init(Rive.kt:138)
      at app.rive.runtime.kotlin.core.Rive$NativeLoader.loadLibrary(Rive.kt:36)
      at app.rive.runtime.kotlin.core.Rive$NativeLoader.loadWithSystemLoader(Rive.kt)
      at java.lang.System.loadLibrary(System.java:1669)
```
  ## Devices / OS versions seen

  Most frequent examples from production:

  - Samsung SM-J415G
  - Samsung SM-A105FN
  - Samsung SM-J737U

  OS versions:

  - Android 9 mostly
  - also Android 8.1.0
  - a few on Android 10 and 12

  One affected event had:

  - ABI/splits: armeabi-v7a
  - OS: Android 9

  ## What we observed

  We compared 10.4.4 and 11.3.1 locally.

  In 10.4.4, the native library load path appears to use ReLinker.

  In 11.3.1, the load path appears to use System.loadLibrary() on Android SDK >= 23.

  We also inspected the packaged armeabi-v7a native lib and noticed that the 11.3.1 binary appears to
  include TLS-related sections / symbols such as __tls_get_addr, while 10.4.4 did not show the same
  pattern.

  This makes it look like a native compatibility regression in the packaged armeabi-v7a library, or a
  loader-path regression, rather than an issue with our usage.

  ## Important note

  Changing renderer does not seem to help.

  The app crashes while loading librive-android.so, before renderer selection matters, so switching
  from RendererType.Rive to another renderer would still hit the same native load failure.

  ## Expected behavior

  Rive.init(...) should not crash the process on supported Android devices when loading the packaged
  native library.

  ## Actual behavior

  Rive.init(...) throws UnsatisfiedLinkError and the app terminates during startup.

  ## Questions

  1. Is 11.3.1 known to have an armeabi-v7a / older Android compatibility issue?
  2. Is this related to TLS relocations or linker compatibility on older Android devices?
  3. Is there a fixed version we should upgrade to?

  If helpful, I can provide more device examples or additional stack traces.

### PR DESCRIPTION
**Pull request by tian000**
## Summary

Add `-femulated-tls` compile option for `armeabi-v7a` in `kotlin/src/main/cpp/CMakeLists.txt`. This prevents the JNI bridge code from generating native TLS relocations (`R_ARM_TLS_DTPMOD32`, type 17) that Android's dynamic linker cannot resolve on API < 29.

## Problem

Since rive-android 11.0.0, the compiled `librive-android.so` for armeabi-v7a contains a `R_ARM_TLS_DTPMOD32` relocation that causes `dlopen` to fail with "unknown reloc type 17" on Android 7–9 (API 24–28). This results in:

- **`UnsatisfiedLinkError`**: JNI methods (e.g., `FileAssetLoader.constructor()`) have no native backing
- **`MissingLibraryException`**: ReLinker cannot find native libraries during initialization

### Production impact (Phantom Wallet, last 7 days)

- **2,831 crash events** across **685 impacted users**
- 93% on Android 7–9, all armeabi-v7a architecture
- Affected devices: Samsung J-series, Huawei Y5/Y6, Infinix, Itel

## Root Cause

The specific source of the TLS relocation is `hydro_random_context` — a 64-byte `__thread` variable in **libhydrogen** (`impl/random.h`), a dependency of the scripting (Luau) feature added in 11.0.0. The `__thread` attribute persists through LTO bitcode compilation and generates a native `R_ARM_TLS_DTPMOD32` in the final shared library.

The primary fix is in the companion rive-runtime PR (rive-app/rive-runtime#93), which:
1. Adds `-femulated-tls` to Premake build options for ARM32
2. Defines `TLS=` (empty) for ARM32 to override libhydrogen's `__thread` macro, converting `hydro_random_context` to a plain static variable (safe: Rive scripting is single-threaded on Android)

This PR adds `-femulated-tls` to the CMake build for the JNI bridge code as a defense-in-depth measure against any future `thread_local` usage.

## Verification

Built `librive-android.so` for armeabi-v7a and compared using `llvm-readelf`:

| Build | R_ARM_TLS_DTPMOD32 | TLS segment | __tls_get_addr |
|---|---|---|---|
| 10.5.1 (known working) | 0 | None | None |
| 11.4.0 (unfixed) | 1 | Yes (64 bytes) | Yes |
| **11.4.0 + fix** | **0** | **None** | **None** |

The fixed build matches the known-working 10.5.1 baseline exactly: zero TLS relocations, zero TLS symbols, no TLS program header.

## Testing

```bash
# Build for armeabi-v7a only
./gradlew :kotlin:assembleRelease -PabiFilters=armeabi-v7a

# Verify no TLS relocations
llvm-readelf -r librive-android.so | grep R_ARM_TLS  # should be empty

# Verify no TLS program header
llvm-readelf -l librive-android.so | grep TLS  # should be empty
```

Install on an ARM32 device running Android 8 or 9 — verify no crash on Rive animation load.

## Related

- Companion: rive-app/rive-runtime#93 (Premake build flags + `TLS=` define)
- Fixes #445
- Fixes #441

### DIFF
--- a/kotlin/src/main/cpp/CMakeLists.txt
+++ b/kotlin/src/main/cpp/CMakeLists.txt
@@ -33,7 +33,10 @@ target_compile_options(rive-android PRIVATE
         -Wall # Enable all warnings
         -fno-exceptions # Disable C++ exceptions
         -fno-rtti # Disable C++ RTTI
-        $<$<CONFIG:Release>:-Oz>) # Optimize for size with -Oz in release
+        $<$<CONFIG:Release>:-Oz> # Optimize for size with -Oz in release
+        # Use emulated TLS on 32-bit ARM to avoid R_ARM_TLS_DTPMOD32 relocations
+        # (type 17) that the Android dynamic linker cannot handle on API < 29.
+        $<$<STREQUAL:${CMAKE_ANDROID_ARCH_ABI},armeabi-v7a>:-femulated-tls>)
 target_compile_definitions(rive-android PRIVATE
         RIVE_ANDROID=
         YOGA_EXPORT=


fabiopinhosword:
Hey team, do you have any idea about when a new release with this fix will be available? We're having some crashes in prod due to this fix.

ErikUggeldahl:
Hi @tian000 (and cc. @fabiopinhosword). Thank you for your efforts investigating and posting a fix for this. I've arrived at the same conclusion as you have as to the root cause, but I needed to adapt the fix at a deeper level, since ultimately the responsible library is libhydrogen, which was included when scripting was added. It required disabling LTO and forcing emulated TLS for that library.

The fix is now in PR and will make it into the next release, presumably 11.4.1. We expect it to go out within a day or so. I'll add you as a co-author for your efforts here. Thanks again!

fabiopinhosword:
Hey @ErikUggeldahl , can you confirm if 11.4.1 has that fix? :) Thank you!

ErikUggeldahl:
@fabiopinhosword Confirmed!

PR #463, opened 2026-07-09
Add getName to experimental ViewModelInstance API

mfazekas:
## Summary

Adds `suspend fun ViewModelInstance.getName()` to the experimental API, returning the editor-assigned instance name resolved by the command server.

Consumers migrating from the classic API currently lose the instance name: `CommandQueue` only exposes `getViewModelInstanceNames(file, viewModelName)` (the per-view-model name list), which cannot tell an existing instance which one it is. Instances created by name echo back what the caller already knew, while default, blank, nested (reference-path) and list-item instances have no way to learn their name at all.

The command queue protocol has no message carrying an instance's own name (`getViewModelInstanceViewModelName` returns the view model's name), so the binding resolves it with `runOnce` on the command server thread via `CommandServer::getViewModelInstance()->name()` and calls back into Kotlin from there — no runtime protocol change needed.

Covered by unit tests and an instrumented test asserting parity with the classic API's `ViewModelInstance.name` for named, default, blank, nested and list-item instances against `data_bind_test_impl.riv`.

The same gap exists in the rive-ios experimental API (RiveRuntime 6.20.4 `ViewModelInstance` has no name member); the same design would apply there for parity.

### DIFF
--- a/kotlin/src/androidTest/kotlin/app/rive/core/ViewModelInstanceNameTest.kt
+++ b/kotlin/src/androidTest/kotlin/app/rive/core/ViewModelInstanceNameTest.kt
@@ -1,0 +1,93 @@
+package app.rive.core
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import app.rive.Artboard
+import app.rive.RiveAndroidTest
+import app.rive.ViewModelInstance
+import app.rive.ViewModelInstanceSource
+import app.rive.ViewModelSource
+import app.rive.runtime.kotlin.core.File
+import app.rive.runtime.kotlin.test.R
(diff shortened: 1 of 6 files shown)


ErikUggeldahl:
Hi @mfazekas. Thanks for this! I've cherry-picked your commit and added it upstream with some enhancements. You should get commit credit when it merges. Just a note if you end up doing enhancements like this in the future: `runOnce` is more of an escape hatch meant to be used for platform specifics (Android in this case) or testing functions. For canonical functions like getting the VMI name, it makes sense to add it to the API surface.

ErikUggeldahl:
Closing as the PR will merge from upstream.

